PR #46, opened 2016-11-18
Features/vagrant lmtp

gdchamal:
Add LMTP service inside the vagrant box, exposing port 4025 to be able to receive mail into caliopen


gdchamal:
it work's at least (some problem with kombu for async delivery that I wll remove for the moment in another PR)

Also a script in devtools to fake a smtp server to talk with caliopen lmtp

PR #49, opened 2016-11-18
Python package for PGP management in backend

gdchamal:
@sapiens-sapide is it ok for you as component structure ?

it's related to previous work: https://trello.com/c/Jxx1Y2iR/272-api-keys-discovery-basics


sapiens-sapide:
OK, thanks !

PR #52, opened 2016-11-21
Import legacy frontend

iamdey:


gdchamal:
LGTM

iamdey:
erf commit author failed. Dunno how to fix author in that case, I restart from scratch, that would be faster

src/frontend/web_application/.gitignore
```diff
@@ -0,0 +1,21 @@
+# dependencies
```
iamdey: Don't you think it could be enough to add "per folder" specific rules ?

iamdey:
new PR #53

PR #58, opened 2016-11-23
return expected response on route /contacts with method POST

sapiens-sapide:
return expected response on route /contacts with method POST


gdchamal:
yes we should use route_url everywhere, my ReSt is not correct on that point

LGTM

PR #59, opened 2016-11-23
set ttl for authentication to 24h for development convenience

sapiens-sapide:
set ttl for authentication to 24h for development convenience


gdchamal:
LGTM, let's merge

PR #60, opened 2016-11-24
Frontend react translate

iamdey:
implement internationalization in frontend


src/frontend/web_application/src/services/i18n.js
```diff
@@ -0,0 +1,73 @@
...
+
+export function getLocale() {
+  let preferedUserLocales = [];
+
+  if (BUILD_TARGET === 'browser') {
+    preferedUserLocales = [
+      document.cookie.locale,
+      ...window.navigator.languages,
+      (window.navigator.language || window.navigator.userLanguage),
+    ].filter(rawLocale => rawLocale);
```
ziir: .filter() here is unnecessary
https://github.com/Gandi/react-translate/blob/master/src/tools/negociateLocale.js#L12

src/frontend/web_application/server/middleware/ssr.js
```diff
@@ -31,10 +31,16 @@ function getMarkup(reactElement, store, assets) {
   ].reduce((str, current) => str.replace(current.key, current.value), tpl);
 }
 
+function applyUserLocaleToGlobal(req) {
+  global.USER_LOCALE = req.cookies.locale || req.locale;
```
ziir: a little bit concerned about that

yanndinendal: Yes, should it be in the `initialState` instead?

yanndinendal: And does `req.locale` really exist? Where does it come from?

src/frontend/web_application/src/services/i18n.js
iamdey: I prefer sanitize myself the array, it's not the purpose of negociateLocale

src/frontend/web_application/server/middleware/ssr.js
iamdey: no it can't be in `initialState` since react-translate does not officially load a locale and a translation after the app boostrap.

req.locale exists, it is a middleware defined in main express file. I don't knw how to be more explicit right now

src/frontend/web_application/src/components/Page/components/Header/components/UserMenu/presenter.jsx
```diff
@@ -45,10 +54,4 @@ class Presenter extends Component {
   }
 }
 
-Presenter.propTypes = {
-  user: PropTypes.shape({
-    name: PropTypes.string.isRequired,
-  }).isRequired,
-};
-
 export default Presenter;
```
iamdey: it is decorated at line 8

PR #66, opened 2016-11-30
Features/dockerize api

gdchamal:
The return of API and CLI in docker containers

### DIFF
--- a/devtools/README.md
+++ b/devtools/README.md
@@ -11,7 +11,8 @@ _For complete documentation, see the [doc/for-developers](../doc/for-developers)
 ## Client development
 
 You will need a running API server where you can consume data, so we have a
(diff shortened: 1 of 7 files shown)


gdchamal:
I don't understand the point with empty line at end of file

For the compose version it should be fixed, also the docker compose stack scenario is a bit described in the README.

iamdey:
I don't knew why for the empty line, but it sounds a pretty good convention: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline#729795

gdchamal:
Files (most of) have an empty line, but it does not appear in github when displaying one of them (caliopen-docker.yaml for example)

iamdey:
Oh, you were right about empty line. Github put a red new line icon when it is missing. I saw some before. well not big deal.

PR #68, opened 2016-12-01
rename routes « thread » to « discussion » at http REST API

sapiens-sapide:
The word « **thread** » is replaced by « **discussion** » in : 

* all HTTP routes of the API (`/discussions`, `/discussions/{discussion_id}`, `/discussions/{discussion_id}/messages`)
* json exchanged with web clients
* py.server
* swagger.json for API documentation

But the word « thread » is still in use in the backend (py.main and py.store) and in the data models (to be changed later).


gdchamal:
nice migration on API side ;-)

LGTM

Issue #62, opened 2016-11-25
route `GET /raws/{raw_id}` doesn't work


sapiens-sapide:
calling this route with a message_id (or a raw_id) returns a 500 « Internal server error » response on client.
Server side, the application console logs : (tail)
```
  File "/Users/stan/Dev/Caliopen/github/Caliopen/src/backend/interfaces/REST/py.server/caliopen_api/message/message.py", line 94, in get
    raw = RawMessage.get(self.user, raw_id)
  File "/Users/stan/Dev/Caliopen/github/Caliopen/src/backend/main/py.main/caliopen_main/message/core/raw.py", line 34, in get
    return super(RawMessage, cls).get(user, key)
  File "/Users/stan/Dev/Caliopen/github/Caliopen/src/backend/main/py.storage/caliopen_storage/core/base.py", line 118, in get
    obj = cls._model_class.get(user_id=user.user_id, **param)
  File "/Users/stan/Dev/Caliopen/github/Caliopen/src/backend/main/py.storage/caliopen_storage/store/model.py", line 45, in get
    raise NotFound(exc)
NotFound
```

(I double checked ids directly into cassandra, so they do exist !)

sapiens-sapide:
this route needs an `external_message_id` to work properly. 
PR #69 proposes to explicitly use the param `ext_msg_id_sum256` (which is the shasum256 of the external message id) to retreive the raw email. If the doc is not found, api returns a 404.

PR #69, opened 2016-12-01
fix incorrect `ressource not found` handling

sapiens-sapide:
route `/raws/` crashed the application if ressource was not found. Now, returns a 404


sapiens-sapide:
This PR fix issue #62 
By the way, it changes the API spec on this route to take a shasum256 as the parameter to retreive the raw doc.

PR #73, opened 2016-12-03
Features/index mapping

gdchamal:
We need to configure indexed document better as everything is analyzed by default.

So an UUID as thread_id, message_id is tokenized.


gdchamal:
I force merge as it's working better than before (lot of bugs around core -> index mapping)

Identify some problems around raw message, fixed in master. So import of mbox work again
as far I tested.

PR #78, opened 2016-12-08
Start a CI using travis with attempt to test for backend and frontend only if needed

gdchamal:
Using git diff on ../src/backend and ../src/frontend to start testing for both components may permit to start having a conditional CI using travis.

@iamdey WDYT ? and if agree can you define frontend testing a minima (not using a functional API for the moment)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ before_install:
  - sudo mv docker-compose /usr/local/bin
 
 script:
-  - docker-compose up -d cassandra elasticsearch redis
-  - docker-compose build
   - devtools/run-tests.sh
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,3 @@ notifications:
     channels:
       - "chat.freenode.net#caliopdev"
     use_notice: true
-addons:
-  sauce_connect:
-    username: AllTheDey
-    access_key:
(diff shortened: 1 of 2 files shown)


devtools/run-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -ev
+
+BACKEND_CHANGE=`git diff ../src/backend`
```
iamdey: I don't understand how this can works, it open diif tool with the diff between HEAD and HEAD.

what do you think of the command I provided: 

```
git diff-tree --no-commit-id --name-only -r HEAD..master src/frontend
```

it shows a list of changed files.  
`master` and `src/frontend` can be defined with var obviously

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+env:
+  COMPOSE_VERSION: 1.9.0
+
+before_install:
+ - curl -L https://github.com/docker/compose/releases/download/${COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+ - chmod +x docker-compose
+ - sudo mv docker-compose /usr/local/bin
+
+script:
+  - docker-compose up -d cassandra elasticsearch redis
```
iamdey: It would be better to not start the whole stack each times. The API of the frontend tests are all mocked and should be fast (tests are run each time a new commit is pushed in a branch/PR).

devtools/run-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+if [[ "x${BACKEND_CHANGE}" != "x" ]]; then
+    echo "##### Doing backend tests"
+    do_backend_tests
+fi
+
+if [[ "x${FRONTEND_CHANGE}" != "x" ]]; then
+    echo "##### Doing frontend tests"
+    do_frontend_tests
+fi
+
```
iamdey: I think an other use case should be implemented: 

* If someone push or merge something on master
* If a tag is pushed 
* If the user launch tests locally

That may implies there is no changes detected on master and so all tests should be started.

devtools/run-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+function do_backend_tests {
+    docker-compose run cli setup
+    docker-compose run cli create_user -e dev@caliopen.local -g John -f Doe -p blablabla
+    docker-compose run cli import -e dev@caliopen.local -f mbox -p /srv/caliopen/code/devtools/fixtures/mbox/dev@caliopen.local
+}
+
+function do_frontend_tests {
+    # XXX to define
+    exit 0
```
iamdey:
```
(cd src/frontend/web_application && npm i && npm test)
```

devtools/run-tests.sh
gdchamal: you're right, in fact missing a git diff <<master>> ../src/backend, but your command is better

.travis.yml
gdchamal: I move storage containers logic only in backend testing

devtools/run-tests.sh
gdchamal: Yes such cases will have to be considered. We will make evolution later to cover them

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+services:
+  - docker
+
+env:
+  COMPOSE_VERSION: 1.9.0
+
+language: node_js
+node_js:
+  - "6"
+  - "6.1"
```
iamdey: this will create a matrix of tests (and run tests for each entry of the matrix). So only 6 is useful actually (may be 4 if we want to support it but there's no reasons for now)

iamdey:
strange:
*  `node stable`, should be node 6 (may be not yet stable should be soon)
* saucelabs fails to connect. I have to check this, maybe comment the configuration until we have no running tests on this platform.

.travis.yml
gdchamal: fixed

PR #79, opened 2016-12-09
Features/pyramid swagger

sapiens-sapide:
requests and responses on py.server MUST validate the swagger schema
all server errors return a verbose error json to client


gdchamal:
This pull request contain too many commits not belonging to the branch, something is not ok with it and it must be rebuild correctly

PR #80, opened 2016-12-09
Features/pyramid swagger2

sapiens-sapide:
py.server enforces requests & responses validation against swagger JSON schema.
py.server returns errors in a verbose json
internal datamodel is consistent with API definitions


src/backend/interfaces/REST/py.server/caliopen_api/user/user.py
```diff
@@ -72,29 +62,14 @@ def collection_post(self):
         self.request.cache.set(user.user_id, tokens)
 
         return {'user_id': user.user_id,
-                'username': user.name,
+                'usernname': user.name,
```
gdchamal: typo (mine sorry)

PR #81, opened 2016-12-10
raw messages have an UUID

sapiens-sapide:
raw message id are generated by the application, no more relying on the external message id.


src/backend/main/py.main/caliopen_main/message/core/raw.py
```diff
@@ -22,19 +23,19 @@ class RawMessage(BaseUserCore):
...
-    _pkey_name = 'raw_id'
+    _pkey_name = 'raw_msg_id'
 
     @classmethod
-    def create(cls, user, message_id, raw):
+    def create(cls, user, raw):
         """Create raw message."""
-        key = hashlib.sha256(message_id).hexdigest()
-        return super(RawMessage, cls).create(user, raw_id=key, data=raw)
+        key = uuid.uuid4()
```
gdchamal: we can avoid generating the raw_msg_id uuid as the store layer have a default value generating it.

In a general manner storage layer is responsible of generating ids, i'm ensure having done that everywhere correctly (Discussion for example)

sapiens-sapide: uuid.uuid4() still in use in : 

* core/contact
* core/thread
* core/message

you cleanup or I cleanup ?

sapiens-sapide: OK, so you're OK to merge this PR ?

gdchamal: yes let's merge ;)

PR #89, opened 2016-12-17
Simplify ESLint config

pascalduez:
1 - Remove redundant config keys 
Most of this is already set by the Airbnb config, and some is outdated.
We'd better let this project maintain that kind of things.

See:
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/es6.js  
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/index.js

2 - Add the `root` key
There's a small performance gain as ESLint won't climb the file system tree to find potential other configs to merge in. Also safer.

See:
http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy

### DIFF
--- a/src/frontend/web_application/.eslintrc
+++ b/src/frontend/web_application/.eslintrc
@@ -5,6 +5,7 @@
   "env": {
     "browser": true,
     "jest": true,
+    "jasmine": true,
     "protractor": true
   },
   "rules": {
--- a/src/frontend/web_application/.eslintrc
+++ b/src/frontend/web_application/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "airbnb",
   "parser": "babel-eslint",
+  "root": true,
   "env": {
     "browser": true,
     "jest": true,
--- a/src/frontend/web_application/.eslintrc
+++ b/src/frontend/web_application/.eslintrc
@@ -1,19 +1,9 @@
 {
   "extends": "airbnb",
   "parser": "babel-eslint",
-  "ecmaFeatures": {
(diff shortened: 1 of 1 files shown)


src/frontend/web_application/.eslintrc
```diff
@@ -1,19 +1,10 @@
...
-    "ecmaVersion": 6,
-    "ecmaFeatures": {
-      "experimentalObjectRestSpread": true
-    }
-  },
+  "root": true,
   "env": {
     "browser": true,
-    "es6": true,
-    "jasmine": true,
```
iamdey: it fails on jasmine dependency, 3 solutions:

* Ignore .spec.jsx? files
* explicit import of jasmine (but style in devdeps so it requires to ignore it)
* re-declare as global

pascalduez: > it fails on jasmine dependency

What do you mean? I haven't noticed any error related to this. 
Also Jest is moving away from Jasmine, they have their own assertion methods now.

pascalduez: Humm you’re right their some errors on Jenkins about it, weird because I checked previously.

pascalduez: You shouldn't call jasmine directly like this I guess. It will be deprecated quite soon, surprising that it works.  
I'll make another PR then ;)

pascalduez:
Added back the `jasmine` key so it does not produce errors after the rebase.

PR #90, opened 2016-12-19
Frontend account - openpgp manager

iamdey:
* list & manage pgpkeys in localstorage
* generate keypair 
* import keypair


src/frontend/web_application/src/components/ContactProfile/index.spec.js
```diff
@@ -1,7 +1,9 @@
 // import contactModule from '../index.js';
 
 describe('component contact-card-summary', () => {
-  it('needs to be implemented', () => {});
+  it('needs to be implemented', () => {
```
yanndinendal: Another way to do this is to use `it.skip(` instead of `it(`, so it will be displayed as not passing in the tests overview at the end of the tests run.

iamdey: sounds good, it tells that how many tests are skipped but not which one

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
```diff
@@ -0,0 +1,150 @@
...
+    expirationTime: publicKey.getExpirationTime(),
+    algorithm: publicKey.primaryKey.algorithm,
+    bitSize: publicKey.primaryKey.getBitSize(),
+    userIds: publicKey.users.map(user => user.userId.userid),
+    keyStatus: Object.keys(keyStatuses)
+      .find(statusLiteral => keyStatuses[statusLiteral] === publicKey.verifyPrimaryKey()),
+  };
+}
+
+@withTranslator()
```
iamdey: If I'm sure this kind of component is not shared, it will be easily moved to "scenes" (aka routed components) But it is not always the case. 
So I don't know how to to this.

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+            <div className="m-openpgp-key__detail">
+              <textarea-field-group
+                label="'openpgp.public-key'|translate"
+                model={publicKeyArmored}
+              />
+            </div>
+            <div className="m-openpgp-key__detail">
+              <textarea-field-group
+                ng-if="!!$ctrl.privateKeyArmored"
```
iamdey: o_O, you right. And `textarea-field-group` is not migrated too. I don't understand what happened, I may missed that or a rebase failure. Thanks!

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
themouette: The only prop you need from translator in this component is `__`, I would ask for it.

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
```diff
@@ -0,0 +1,150 @@
...
+    this.toggleDetails = this.toggleDetails.bind(this);
+    this.handleDeleteKey = this.handleDeleteKey.bind(this);
+    this.keyStatus = enums.keyStatus;
+  }
+
+  componentWillMount() {
+    this.initViewValues();
+  }
+
+  initViewValues() {
```
themouette: By the way, the only way you can be sure the component has the right data rendered is to provoke a rerender.

You can do this by changing state or props.

By setting a component property (`this.viewValues`), there is no guarantee over time the render will be performed with the updated value.

src/frontend/web_application/src/components/ContactProfile/index.spec.js
themouette: It tells which tests were skipped as well:

screenshot from 2016-12-20 10-34-24

themouette:
oh, forgot you were using jest !

src/frontend/web_application/src/scenes/Account/components/AccountOpenPGPKeys/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+const mapDispatchToProps = dispatch => ({
+  onDeleteKey: ({ fingerprint }) => {
+    dispatch(openPGPKeychain.deleteKey(fingerprint));
+  },
+  onImportKey: ({ publicKeyArmored, privateKeyArmored }) => {
+    dispatch(openPGPKeychain.importKeyPairChains(publicKeyArmored, privateKeyArmored));
+  },
+  onGenerateKey: ({ name, email, passphrase }) => {
+    dispatch(openPGPKeychain.generate(name, email, passphrase));
```
ziir: You could keep the same signature as your action-creators', their return value, reducing boilerplate code and avoiding typos by using redux' `bindActionCreators` https://github.com/reactjs/redux/blob/master/src/bindActionCreators.js

```js
import { bindActionCreators } from 'redux';

const mapDispatchToProps = dispatch => ({
  onDeleteKey: openPGPKeychain.deleteKey,
  ...
});
```

src/frontend/web_application/src/scenes/Account/components/AccountOpenPGPKeys/presenter.jsx
```diff
@@ -0,0 +1,117 @@
...
+            server and you will not able to see it on any other devices.
+          </p>
+          <p>
+            Be warned, the key pair generation is pretty slow and will freeze this page for
+            approximatively 20 seconds (depends on your device capacities). A fix is in progress but
+            may takes time to become available.
+          </p>
+        </div>
+        {
+          privateKeys.map((keyPair, key) => (
```
ziir: Both for performance and readability, some people including me consider not defining functions in render method a fairly good practice - you could have a `AccountOpenPGPKeysInstance.renderPrivateKey` method and do `{privateKeys.map(this.renderPrivateKey)}`

ziir: Maybe relevant reading on the topic https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f

src/frontend/web_application/src/services/openpgp-keychain-repository/local-storage-helper.js
```diff
@@ -0,0 +1,20 @@
+export function findAll(namespace) {
+  return Object.keys(window.localStorage)
```
ziir: localStorage is not always availalble, you'll probably be safer testing for it's support first, or you'll end up with a TypeError ;)

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      showDetails: false,
+    };
+
+    this.toggleDetails = this.toggleDetails.bind(this);
+    this.handleDeleteKey = this.handleDeleteKey.bind(this);
+    this.keyStatus = enums.keyStatus;
```
ziir: Why assigning it to the instance?

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
iamdey: FYI the doc indicates to not update state here but in `componentWillReceiveProps` https://facebook.github.io/react/docs/react-component.html#componentwillupdate

src/frontend/web_application/src/scenes/Account/components/AccountOpenPGPKeys/index.js
iamdey: :100: I added a comment for refactoring

src/frontend/web_application/src/services/openpgp-keychain-repository/local-storage-helper.js
iamdey: added a test in top of file. The following code is only be present on browser built

src/frontend/web_application/src/components/OpenPGPKey/index.jsx
iamdey: I was not proud of that; I always keep in mind the quote from Head first design pattern book "program an interface, not an implementation" but since there are no dependency injection inside react components it is just weird.

Well I fixed with @themouette reviews.

src/frontend/web_application/src/services/openpgp-keychain-repository/local-storage-helper.js
ziir: Ok, but localStorage is also not available in all the browsers

PR #101, opened 2017-01-09
Prevent using cache when cloning in Dockerfile

iamdey:
+ fix docker import command (& doc)


gdchamal:
Well will do the job for the moment until a better solution can be found

PR #106, opened 2017-01-11
Features/api tags

gdchamal:
Base API to list and manage user tags:
- GET /tags
- POST /tags
- DELETE /tags/<name>
- GET /tags/<name>

Output of collection will contain system tags:

```
In [11]: res = requests.get('%s/tags' % (base_url), auth=auth, headers=headers)

In [12]: pprint(res.json())
{u'count': 3,
 u'tags': [{u'name': u'INBOX', u'type': u'system'},
           {u'name': u'SPAM', u'type': u'system'},
           {u'name': u'IMPORTANT', u'type': u'system'}]}
```

### DIFF
--- a/doc/api/swagger.json
+++ b/doc/api/swagger.json
@@ -549,7 +549,6 @@
           }
         ],
         "consumes": [
-          "application/x-www-form-urlencoded",
           "application/json"
         ],
         "parameters": [
--- a/doc/api/swagger.json
+++ b/doc/api/swagger.json
@@ -18,20 +18,29 @@
         ],
         "security": [],
         "consumes": [
-          "application/x-www-form-urlencoded"
+          "application/json"
         ],
         "parameters": [
           {
-            "name": "username",
-            "in": "formData",
-            "type": "string",
-            "required": true
-          },
-          {
-            "name": "password",
-            "in": "formData",
-            "type": "string",
-            "required": true
+            "name": "authentication",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "type": "object",
+              "properties": {
+                "username": {
+                  "type": "string"
+                },
+                "password": {
+                  "type": "string"
+                }
+              },
+              "required": [
+                "username",
+                "password"
+              ],
+              "additionalProperties": false
+            }
           }
         ],
         "produces": [
@@ -540,7 +549,8 @@
           }
         ],
         "consumes": [
-          "application/x-www-form-urlencoded"
+          "application/x-www-form-urlencoded",
+          "application/json"
         ],
         "parameters": [
           {
@@ -1026,30 +1036,6 @@
         }
       },
       "delete": {
-        "description": "Not Yet Implemented",
-        "tags": [
-          "users"
-        ],
-        "security": [
-          {
-            "basicAuth": []
-          }
-        ],
-        "parameters": [
-          {
-            "name": "user_id",
-            "in": "path",
-            "required": true,
-            "type": "string"
-          }
-        ],
-        "responses": {
-          "default": {
-            "description": "route not implemented, should raise an error"
-          }
-        }
-      },
-      "put": {
         "description": "Not Yet Implemented",
         "tags": [
           "users"
--- a/doc/api/swagger.json
+++ b/doc/api/swagger.json
@@ -4375,6 +4375,9 @@
             "basicAuth": []
           }
         ],
+        "produces": [
+          "application/json"
+        ],
         "responses": {
           "200": {
             "description": "Tags returned",
--- a/doc/api/swagger.json
+++ b/doc/api/swagger.json
@@ -1611,95 +1611,6 @@
         }
       }
     },
-    "/tags": {
-      "get": {
-        "description": "Returns tags visible to current user according to given parameters",
-        "tags": [
-          "tags"
-        ],
-        "security": [
-          {
-            "basicAuth": []
-          }
-        ],
-        "responses": {
-          "200": {
-            "description": "Tags returned",
-            "schema": {
-              "type": "object",
-              "properties": {
-                "total": {
-                  "type": "integer",
-                  "format": "int32",
-                  "description": "number of tags found for user for the given parameters"
-                },
-                "contacts": {
-                  "type": "array",
-                  "items": {
-                    "type": "object",
-                    "properties": {
-                      "type": {
-                        "type": "string"
-                      },
-                      "label": {
-                        "type": "string"
-                      },
-                      "name": {
-                        "type": "string"
-                      }
-                    },
-                    "required": [
-                      "name",
-                      "type"
-                    ],
-                    "additionalProperties": false
-                  }
-                }
-              }
-            }
-          }
-        }
-      },
-      "post": {
-        "description": "Create a new Tag for an user",
-        "tags": [
-          "tags"
-        ],
-        "security": [
-          {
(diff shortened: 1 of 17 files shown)


doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+            "description": "Tags returned",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "total": {
+                  "type": "integer",
+                  "format": "int32",
+                  "description": "number of tags found for user for the given parameters"
+                },
+                "contacts": {
```
iamdey: I don't understand what is the purpose of "contacts"

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+        "tags": [
+          "tags"
+        ],
+        "security": [
+          {
+            "basicAuth": []
+          }
+        ],
+        "consumes": [
+          "application/x-www-form-urlencoded"
```
iamdey: json ?

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+        "produces": [
+          "application/json"
+        ],
+        "responses": {
+          "200": {
+            "description": "Tags returned",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "total": {
```
iamdey: `total` sounds not useful for now https://projects.invisionapp.com/share/X59YPM1GC#/screens/213687066

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+                  "type": "integer",
+                  "format": "int32",
+                  "description": "number of tags found for user for the given parameters"
+                },
+                "contacts": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "type": {
```
iamdey: what are allowed types ? (enum ?)

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+                        "type": "string"
+                      },
+                      "label": {
```
iamdey: is `label` the translation of `name` ? 

btw, don't you need to declare an id ?

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+            "basicAuth": []
+          }
+        ],
+        "consumes": [
+          "application/x-www-form-urlencoded"
+        ],
+        "parameters": [
+          {
+            "name": "name",
+            "in": "formData",
```
iamdey: what is `in` ?

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+            "type": "string",
```
iamdey: enum ?

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+        "produces": [
+          "application/json"
+        ],
+        "responses": {
+          "200": {
+            "description": "User tag creation completed",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "location": {
```
iamdey: `location` ?

doc/api/swagger.json
```diff
@@ -4363,6 +4363,216 @@
...
+                  "type": "string",
+                  "description": "url to retrieve new tag's infos at /tags/{name}"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/tags/{name}": {
```
iamdey: an `id` can be more easily used ? (in case the name doesn't use standard characters)

src/backend/defs/rest-api/paths/tags.json
```diff
@@ -0,0 +1,146 @@
...
+                "type": "string",
+                "description": "url to retrieve new tag's infos at /tags/{name}"
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  "tags_{name}": {
```
iamdey: I'm not sure it is very useful to get only one tag.

src/backend/defs/rest-api/swagger-root.json
```diff
@@ -19,7 +19,9 @@
     "/discussions": {"$ref": "paths/discussions.json#/discussions"},
     "/discussions/{discussion_id}": {"$ref": "paths/discussions.json#/discussions_{discussion_id}"},
     "/discussions/{discussion_id}/messages": {"$ref": "paths/discussions.json#/discussions_{discussion_id}_messages"},
-    "/raws/{raw_msg_id}": {"$ref": "paths/raws.json#/raws_{raw_msg_id}"}
+    "/raws/{raw_msg_id}": {"$ref": "paths/raws.json#/raws_{raw_msg_id}"},
+    "/tags": {"$ref": "paths/tags.json#/tags"},
+    "/tags/{name}": {"$ref": "paths/tags.json#/tags_{name}"}
```
iamdey: in near futur I will need to list tags related to entites, there is two (or three) options:

- the entity gives a collection of tags
- a sub-route of an entity gives the collection
- each tags gives a list of associated entities

doc/api/swagger.json
gdchamal: I add it to be consistent with others API. But for configuration API may we can decide to use a simpler model (direct list of results)

doc/api/swagger.json
gdchamal: name is the ID and yes label is for translation purposes

src/backend/defs/rest-api/paths/tags.json
gdchamal: As POST reply with the location of the created tag it's for consistency.

src/backend/defs/rest-api/swagger-root.json
gdchamal: Entities will produce list of associated tags, but they can produce only name of them or complete tag structure.

['important', 'boulot']

[{'name': 'important', 'type': 'system', 'label': 'Important', ....]

The third point is a search in some entities having this tag.

Remark: a tag on an entity may does not exist anymore in the current list on GET /tags. Managing a list of current tags and have historical tag values differents is 2 distinct things

doc/api/swagger.json
gdchamal: I use name as primary key, may it's not a correct design and we can use an UUID as everywhere else.

As it's a configuration API, I thing this can be simpler this way. We have to deal correctly with HTML encoding

doc/api/swagger.json
gdchamal: it's the way to declare the reference to the input parameter structure

doc/api/swagger.json
gdchamal: bad copy paste ;)

doc/api/swagger.json
gdchamal: I begin in swagger world, I will check about enum restriction

doc/api/swagger.json
gdchamal: something is not ok it should be only json as we see on the POST /authentications problem.

I'll update

doc/api/swagger.json
gdchamal: where does the data come from (querystring, body, formulaire)

gdchamal:
I pushed update. Missing patch route, but as a Tag is only a name that can be define by an user update of a tag is not really needed. DELETE + POST will do the job ;)

PR #107, opened 2017-01-11
proposal for the username format rules and validation.

sapiens-sapide:
proposal for the username format rules and validation.

### DIFF
--- a/doc/RFCs/username_specifications.md
+++ b/doc/RFCs/username_specifications.md
@@ -4,7 +4,7 @@ This document describes valid **username** for the creation of an account within
 
 #### Preambule
 Caliopen's users should be able to take (almost) any username they want to, as long as this username could make the local-part of an email address.  
-Despite the very open RFC 5322 specifications for the Internet Message Format, that authorizes — for years — a broad range of characters within the local-part of an email address, the sad reality is that many of Mail Transfer Agents still refuse to handle emails with "special" characters in their address. Consequently, we adopted a more restrictive specification than 5322 for the username.
+The current specification is based on RFC 5322 for the Internet Message Format, but it does not allow "comment" and "quoted-string" lexical tokens. Consequently, some special characters are not allowed (see below).
 
 ##### Definition :
 An **username** is the unique identifier an user makes use of to create an account within a Caliopen instance. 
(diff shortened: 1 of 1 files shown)


doc/RFCs/username_specifications.md
```diff
@@ -0,0 +1,84 @@
...
+An **username** is the unique identifier an user makes use of to create an account within a Caliopen instance. 
+Username is only an identifier for the user's account : it is not necessarily the user's real name, or email, or nickname… It will only be used as a credential for the purpose of identifying the user when logging in Caliopen. The username is unique within a Caliopen instance.
+
+NB : once an user has chosen an _username_, she/he will be able to create or add some « identities », that are made of : first name, family name, email, etc.
+
+##### Format :
+ 
+* Username are made of utf-8 "characters".  
+By "character" we mean a single Unicode grapheme, encoded as a single Unicode code point.
+* Username **is** case sensitive.
```
gdchamal: the username must definitly be case insensitive to avoid identity scam/squatting and so

sapiens-sapide: I think you make a confusion between the **username**, that is part of the user's credential, and the fact that the username **may** be used to form the local part of an email. So, as a credential, the username **must** be case sensitive, right ?

sapiens-sapide:
In a general way, I think there is a misunderstanding of what the _username_ is for. It is a credential, nothing more.  
Caliopen is a general messaging platform. An user could be a Caliopen registred member, with an username, and do not have any email account connected to the platform. Only her/his twitter and facebook accounts for example.  
Thus, I definitely think the _username_ should have nothing in common with the users's messaging accounts (email, twitter, irc, etc.)

iamdey:
agree with @sapiens-sapide, the username might help to generate a default local part the instance user email. But we should let the user change it. I know it represents a certain amount of work.

doc/RFCs/username_specifications.md
LaurentChemla: We've always thought that the **username** really was the local part of an email, even if the user does'nt know it at first (he may choose to use it later when he becomes aware that his other social identities have poor PIs). I'm not sure we really want to change our mind about it, as it may be too confusing for an average user to create an email specific to the Caliopen instance that would be different from his **username**.

This doesn't mean that username must me case insensitive, as local part of an email **may** be case sensitive, but AFAIK no major emailing system distinguish different addresses based on case. So I would prefer those usernames to be case insensitive anyway.

myleneb:
as far as I understand, I fully agree with @sapiens-sapide

gdchamal:
A last detail, no commit must be named wip. Merge this PR in squash mode

PR #109, opened 2017-01-13
Feature/backend/new smtp system

sapiens-sapide:
This branch merge 4 development branches that : 
* creates a smtpd
* delivers incoming email via NATS messaging system
* introduces the very first GO components within Caliopen backend.


src/backend/protocols/go.smtp/mocks/mapTableCassandra.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+	"github.com/gocassa/gocassa"
+	"time"
+)
+
+// This test assumes that cassandra is running on default port locally and
+// that the keySpace called 'test' already exists.
+
+type Sale struct {
```
gdchamal: what does it do here ?

gdchamal:
Also a general gofmt before landing in master is needed

PR #111, opened 2017-01-17
Frontend signup

myleneb:
This create new raw components:
```
<SignupForm /> 
<Brand /> 
<Title /> 
<PasswordStrenght />
```
and update some styles to fit UI for account creation.


myleneb:
`<SignupForm />`

SignupForm preview

src/frontend/web_application/src/components/form/PasswordStrenght/style.scss
```diff
@@ -0,0 +1,75 @@
...
+    padding-left: 30px;
+    font-size: $co-font__size--small;
+    font-weight: 700;
+    text-shadow: 0 0 1px $co-color__shadow;
+    text-transform: uppercase;
+  }
+
+  &__graph {
+    @include flex-grid-row;
+    padding-left: 0 !important;
```
iamdey: this might not be here (it looks like a bug mixin row and column and not using `flex-grid-size`)

src/frontend/web_application/src/components/form/PasswordStrenght/style.scss
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  i {
```
iamdey: -> BEM `&__icon`

src/frontend/web_application/src/components/form/PasswordStrenght/style.scss
```diff
@@ -0,0 +1,75 @@
...
+    @include flex-grid-column(shrink);
+  }
+
+  &__guide {
+    @include flex-grid-column;
+    height: 3px;
+    margin-top: 5px;
+    padding-right: 0 !important;
+    padding-left: 0 !important;
```
iamdey: `flex-grid-size`

src/frontend/web_application/src/components/form/PasswordStrenght/style.scss
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  i {
+    @include flex-grid-column(shrink);
+  }
+
+  &__guide {
+    @include flex-grid-column;
+    height: 3px;
+    margin-top: 5px;
```
iamdey: hmmm -> config ?

src/frontend/web_application/src/components/form/PasswordStrenght/style.scss
```diff
@@ -0,0 +1,75 @@
...
+    padding-right: 0 !important;
+    padding-left: 0 !important;
+    background-color: $co-color__fg__back--lower;
+  }
+
+  &__bar {
+    height: 100%;
```
iamdey: `$co-component__height` ? or something like that ?

src/frontend/web_application/src/components/form/Switch/index.jsx
```diff
@@ -2,26 +2,30 @@ import React, { PropTypes } from 'react';
...
+          type="checkbox"
+          className="m-switch__input"
+          id={id}
+          {...inputProps}
+        />
+        <label className="m-switch__paddle" htmlFor={id}>
+          <span className="show-for-sr">{label}</span>
+        </label>
+      </div>
+      { duplicateLabel && <label htmlFor={id} className="m-switch__label">{label}</label> }
```
iamdey: I would prefer a `SwitchFieldGroup` component instead. Here we have an Element outside of the Block

src/frontend/web_application/src/components/form/Switch/style.scss
```diff
@@ -16,5 +16,12 @@ $switch-height: 2rem !default;
 
   &__paddle {
     @include switch-paddle;
+    background-color: $co-color__fg__back--lower;
+  }
+
+  &__label {
+    padding-left: $co-grid__gutter--small;
+    font-size: .9rem;
```
iamdey: Does this might look weird in a form with multiple label etc. ?

Also may be we need to add font size modifiers to TextBlock, something like `<label><TextBlock inline small>bla</TextBlock></label>`. WDYT ?

src/frontend/web_application/src/components/form/TextFieldGroup/index.jsx
```diff
@@ -5,18 +5,28 @@ import { InputText, FieldErrors } from '..';
...
-  label, errors = [], expanded = true, className, showLabelforSr = false, ...inputProps
+  label,
+  errors = [],
+  expanded = true,
+  className,
+  showLabelforSr = false,
+  ...inputProps
 }) => {
   const id = uuidV1();
+  let errorClass = false;
```
iamdey: shortly:

```
const errorClass = errors.length > 0;
```

(I prefer hasError, ;) )

src/frontend/web_application/src/components/form/TextFieldGroup/style.scss
```diff
@@ -1,11 +1,24 @@
...
   }
 
   &__errors {
-    margin-top: map-get($co-form__spacing, 'medium');
+    position: relative;
+    margin-top: -(map-get($co-form__spacing, 'medium'));
   }
 }
+
+.m-field-errors {
```
iamdey: there is already a FieldErrors component with its styles.

this might introduce bugs in styles : 
it is loaded only if a form contains a TextFieldGroup component and will be applied to all other errors

src/frontend/web_application/src/components/form/TextFieldGroup/style.scss
```diff
@@ -1,11 +1,24 @@
 @import '../../../styles/common';
 
 .m-text-field-group {
+  position: relative;
```
iamdey: I don't understand what is the purpose of those position directives

src/frontend/web_application/src/styles/config/config.scss
```diff
@@ -11,7 +11,8 @@ $co-component__spacing: .5rem;
 $co-form__width: (
   small:  10rem,  // 160 px
   medium: 13rem,  // 208 px
-  large:  16rem   // 256 px
+  large:  16rem,   // 256 px
+  xlarge: 320px,
```
iamdey: 20rem according to http://www.remcalculator.com/

src/frontend/web_application/src/components/form/FieldErrors/style.scss
```diff
@@ -2,9 +2,10 @@
 @import '../../../styles/vendor/bootstrap_foundation-sites';
```
iamdey: It seems it's not used anymore

src/frontend/web_application/src/components/form/FieldErrors/style.scss
```diff
@@ -2,9 +2,10 @@
...
 
 .m-field-errors {
-  @include callout($color: $co-color__alert);
```
iamdey: :+1:

src/frontend/web_application/src/components/TextBlock/index.jsx
```diff
@@ -2,16 +2,28 @@ import React, { PropTypes } from 'react';
...
+
+  return (
+    <span className={textBlockClassName} {...props} />
+  );
+};
 
 TextBlock.propTypes = {
   inline: PropTypes.bool,
   className: PropTypes.string,
+  size: PropTypes.string,
```
iamdey: :+1: 

```
size: PropTypes.oneOf(['small']),
weight: PropTypes.oneOf(['strong']),
```

src/frontend/web_application/src/components/form/Switch/index.jsx
```diff
@@ -1,31 +1,24 @@
...
+    <label className="m-switch__paddle" htmlFor={id}>
+      <span className="show-for-sr">{label}</span>
+    </label>
+  </div>
+);
 
 Switch.propTypes = {
   label: PropTypes.string.isRequired,
-  duplicateLabel: PropTypes.bool,
+  id: PropTypes.string.isRequired,
```
iamdey: may be: 

```
  id: PropTypes.string

// and 

const checkboxId = id || uuidV1();
```

So it's optional and we are still able to use it without the group

iamdey: Oh okay, we have to use SwitchFieldGroup now. 

So you need to look for Switch component in the project and replace it by SwitchFieldGroup, else it will be "broken".

src/frontend/web_application/src/components/form/TextFieldGroup/style.scss
```diff
@@ -1,11 +1,24 @@
...
+
   &__label {
     display: block;
   }
 
   &__errors {
-    margin-top: map-get($co-form__spacing, 'medium');
+    position: relative;
+    margin-top: -(map-get($co-form__spacing, 'medium'));
```
iamdey: hmm there's a problem here: we shouldn't have to remove a padding.

I suppose it's my mistake, I did say a component should not have margin because it's more difficult to remove than add it. But I didn't do it there.

So it's the same remark as previous. the padding is fixed here but not for other form components.

I propose to not fix this margin here. Merge this PR. Then create a new one that fix the spacing for all cases

src/frontend/web_application/src/components/form/TextFieldGroup/style.scss
iamdey: still useful ?

src/frontend/web_application/src/components/form/Switch/style.scss
iamdey: still used ?

src/frontend/web_application/src/components/form/TextFieldGroup/style.scss
```diff
@@ -1,11 +1,13 @@
...
 .m-text-field-group {
+  position: relative;
+
   &__label {
     display: block;
   }
 
   &__errors {
-    margin-top: map-get($co-form__spacing, 'medium');
+    margin-top: -(map-get($co-form__spacing, 'medium'));
```
iamdey: hmm there's a problem here: we shouldn't have to remove a padding.

I suppose it's my mistake, I did say a component should not have margin because it's more difficult to remove than add it. But I didn't do it there.

So it's the same remark as previous. the padding is fixed here but not for other form components.

I propose to not fix this margin here. Merge this PR. Then create a new one that fix the spacing for all cases

PR #116, opened 2017-01-19
Front/ <LoginForm> and <AuthPage> styles

myleneb:
Make `<LoginForm>` and `<AuthPage>` fit the UI

### DIFF
--- a/src/frontend/web_application/src/components/LoginForm/style.scss
+++ b/src/frontend/web_application/src/components/LoginForm/style.scss
@@ -1,0 +1,26 @@
+@import '../../styles/common';
+@import '../../styles/vendor/bootstrap_foundation-sites';
+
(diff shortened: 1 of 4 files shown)


src/frontend/web_application/src/components/LoginForm/style.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../styles/common';
+@import '../../styles/vendor/bootstrap_foundation-sites';
+
+
+.s-login {
+  max-width: map_get($co-form__width, 'xlarge');
+  margin: auto;
```
iamdey: I'm not sure about this. A component should not position itself. It's the parent purpose (like margin).

for example, `&__brand` is done right in my mind

PR #117, opened 2017-01-20
Design/go api/application scaffolding

sapiens-sapide:
* go rest http api server
* go proxy for routing api/v1 & api/v2
* serving route /api/v2/username/isAvailable


src/backend/main/go.main/cmd/caliopen/cli-cmds/APIs.go
```diff
@@ -0,0 +1,206 @@
...
+		log.WithError(err).Fatalf("Could not parse config file: <%s>", cmdConfig.Apis.GoRESTconfigFile)
+	}
+
+	err = rest_api.InitializeServer(apiConfig)
+	if err != nil {
+		log.Fatal(err)
+	}
+	go rest_api.StartServer()
+
+	if withPyramid {
```
gdchamal: nope it's not the role of an http server to start anoher process. There is enough OS tooling for that

sapiens-sapide: I'll remove that bootstraping, but notice that go.main is not an « http server ».  
In my mind, it is the « master of ceremony », ie the binary that manages the whole Caliopen's instance : reads the configuration file and exports the settings to other processes, start services/processes, monitor them (keeps all the other processes running. If a child process dies, another one is restarted automatically),  is the centralized logging facility, etc…  
But let's forget it for now.

PR #119, opened 2017-01-20
Features/contact tags

gdchamal:
Make possible to manage tags on Contact using a full structure {'type': '', 'name': '', 'tag_id':}
with PATCH method


gdchamal:
yes tag type is not an user input. it's system for ones loaded on user creation, it's user for those created by an user

PR #122, opened 2017-01-24
Features/device base

gdchamal:
Bring basic device storage capabilities and a CRUD interface for user devices

### DIFF
--- a/src/backend/defs/rest-api/objects/NewDevice.json
+++ b/src/backend/defs/rest-api/objects/NewDevice.json
@@ -1,0 +1,18 @@
+{
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "type": {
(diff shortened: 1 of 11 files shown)


src/backend/defs/rest-api/objects/NewDevice.json
```diff
@@ -0,0 +1,18 @@
+{
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "type": {
```
iamdey: may be an enum ? 
does type mean [desktop|smartphone|tablet|watch|washing machine] ?

later it would be nice to have OS and version.

gdchamal: set an enum with these values, default tu unknow one if not provided by client

PR #125, opened 2017-01-26
api POST /users for user creation

sapiens-sapide:
api POST /users for user creation
+ bug fixes in py caliopen objects algorithms

### DIFF
--- a/doc/api/swagger.json
+++ b/doc/api/swagger.json
@@ -78,7 +78,8 @@
                     }
                   }
                 }
-              }
+              },
+              "additionalProperties": false
             }
           },
           "401": {
@@ -214,6 +215,9 @@
                           "is_primary": {
                             "type": "boolean"
                           },
+                          "is_backup": {
+                            "type": "boolean"
+                          },
                           "label": {
                             "type": "string"
                           },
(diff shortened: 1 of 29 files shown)


doc/api/swagger.json
```diff
@@ -2032,7 +1741,28 @@
...
-                    "type": "object"
+                    "type": "object",
+                    "properties": {
+                      "address": {
+                        "type": "string"
+                      },
+                      "is_primary": {
+                        "type": "boolean"
+                      },
+                      "is_backup": {
```
iamdey: may be I don't understand well : It looks like the recover email is set in contact details of the user.
I am not sure it is good idea:

 - a user can delete a contact details ... the recovery emal as well
 - so we both need to add a constraint on deletion and create the engine to change the recovery email etc. (more code = more bug)
 - it also requires more UX reflexion to manage this kind of cases.
 - a simple contact has no is_backup concept
 - right now we can't store a confirmation or validation of this email

I've an other proposition: add a "custom" field to the user : the recovery_email and the date of the validation; (it might be usefull to add the history of this field).
On user creation we can simply duplicate the recover_email into contact_details

gdchamal:
An user can't delete is own contact information so backup email if we keep such design.

But you're right user contacts do not have such email is_backup information. So I revert design we discussed too quicly this afternoon with @sapiens-sapide.

We must in fact add such email backup information to User object and not its related Contact one.

gdchamal:
I wasn't ensure about my decision. I'm now certain I'd got a wrong one ;)

sapiens-sapide:
commit #ce9f976 makes use of a «recovery_email» within User object, no more «is_backup» flag within Contact.

PR #129, opened 2017-01-27
Frontend: Devices management

myleneb:
Creates first draft for DevicesManagement component in Settings scene.


src/frontend/web_application/src/scenes/Settings/components/DevicesManagement/components/DisplayDevice/style.scss
```diff
@@ -0,0 +1,35 @@
+@import '../../../../../../styles/common';
+@import '../../../../../../styles/vendor/bootstrap_foundation-sites';
+
+.m-device {
+
+  &__row {
```
iamdey: I'm asking myself if it should be in FormRow directly with a modifier, something like `<FormRow gutter />` which add ... gutters

src/frontend/web_application/src/scenes/Settings/components/DevicesManagement/components/DisplayDevice/style.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding: map-get($co-form__spacing, 'small') map-get($co-form__spacing, 'large');
+
+    &--separated {
+      border-top: 1px solid $co-color__bg__back;
+    }
+  }
+
+  &__title {
+    padding: map-get($co-form__spacing, 'large') 0;
```
iamdey: `padding-top & padding-bottom`

src/frontend/web_application/src/scenes/Settings/components/DevicesManagement/components/DisplayDevice/style.scss
myleneb: I agree, I can improve that in a future PR

PR #135, opened 2017-02-01
Refactoring go code

sapiens-sapide:


src/backend/interfaces/REST/go.server/api_server.go
```diff
@@ -77,37 +77,35 @@ func (server *REST_API) start() error {
...
-        }
-	return err
-}
-
-func BindBackend() gin.HandlerFunc {
-	return func(c *gin.Context) {
-		c.Set("APIStore", server.store)
-		c.Next()
+	if err != nil {
+		log.WithError(err).Info("unable to start gin server")
```
gdchamal: don't log info level if it's an error, warn at least

It's a generic remark, you sometimes log an error using info level

PR #136, opened 2017-02-01
fix response server on route /me (was not compliant to swagger definition).

sapiens-sapide:
fix response server on route /me (was not compliant to swagger definition).


gdchamal:
LGTM

PR #137, opened 2017-02-02
Feature/backend/message tags

gdchamal:
This branch permit to create a message with a list of ResourceTag structure (tag_id,type, name) only backend side, and update tags using PATCH method on /discussions/<discussion_id>/messages/<message_id> work.

I'm a bit ensure about such second level route that need discussion_id reference for individual messages management actions and may a simplest route (/messages/...) is better. discussion_id is a filtering option between others for others searches in messages

We will surely need message action route (PREFIX/messages/<message_id>/action {'action':  something}) somehow (mark as read/unread, send, ... can be such actions).

Such action are single event and do not fit PATCH semantic (but technically can) from my point of view and need a dedicated action route.

To be discussed for API routing question (second level resource and action specific route)


sapiens-sapide:
On my side, I prefere the 'messages' route, ie : `POST` on `PREFIX/messages/<message_id>/action`, with a JSON describing action to complete.
@iamdey ?

iamdey:
I'm ok for the simplistic route. I prefer not to add complexity just to retrieve the `discussion_id` required by the route

gdchamal:
I'll merge in current status and will do a complete proposal for new messages routes
using simple collection hierarchy

Issue #145, opened 2017-02-02
export the logo in svg without margins


iamdey:
We need to remove png from the client app : but the current Brand component has a svg with margins and cannot be used easily

![capture d ecran de 2017-02-02 16 51 31](https://cloud.githubusercontent.com/assets/65737/22556631/2896df2c-e968-11e6-839d-4c3e1475f641.png)

@steventlucek can you re-export the logo ?

---

* [ ] export svg - steven
* [ ] update component - dey/mylene
* [ ] replace png by svg in header - dey/mylene
* [ ] fix height of Brand component in signin/signup  interfaces - dey/mylene

PR #147, opened 2017-02-02
Front: Logo in svg without margins

myleneb:
Updates logo svg (removes margins) and related componnents (Header, Auth page)


iamdey:
fix #145

PR #154, opened 2017-02-08
Refactor/frontend/page

iamdey:
move Page from components directory to layouts


src/frontend/web_application/src/layouts/Page/style.scss
```diff
@@ -0,0 +1,20 @@
...
+
+@import '../../styles/common';
+
+.l-body {
+  &__content {
+    // XXX: need a better fix for call-to-action override on mobile
+    margin-bottom: 160px;
+    padding-top: 45px;
+
+    background-color: $co-color__fg__back;
```
myleneb: maybe `$co-color__bg__back` instead, to fit UI?

iamdey: I prefer not here. We need the background to fit all use cases (Settings, Discussions, Contacts etc.)

PR #155, opened 2017-02-08
Migrate StickyNavBar component

iamdey:
capture d ecran de 2017-02-08 18 18 42

It also adds a little css animation on the navbar on scroll wich was present on previous poc

### DIFF
--- a/src/frontend/web_application/src/layouts/Page/components/Navigation/components/StickyNavBar/index.jsx
+++ b/src/frontend/web_application/src/layouts/Page/components/Navigation/components/StickyNavBar/index.jsx
@@ -1,0 +1,40 @@
+import React, { Component, PropTypes } from 'react';
+import classnames from 'classnames';
+
+class StickyNavbarClass extends Component {
+  static propTypes = {
+    className: PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/frontend/web_application/src/layouts/Page/components/Navigation/components/StickyNavBar/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.handleScroll);
+  }
+
+  handleScroll(ev) {
+    const scrollSize = ev.srcElement.documentElement.scrollTop || window.scrollY;
+
+    this.setState({ isSticky: scrollSize > 10 });
```
ziir: You might want to debounce these scroll events, as you're effectively triggering reflow + repaints + react setState + react render on every scroll event

iamdey: you are right, it's good as it is. Let's talk in the PR #162

PR #157, opened 2017-02-10
Feature/protocols/smtp outbound

sapiens-sapide:
listen to nats messages on topic «outboundSMTP» to build and send email from a message draft.


gdchamal:
LGTM and should break nothing in the current stack

PR #160, opened 2017-02-13
Feature/frontend: ContactBook

myleneb:
Add ContactBook layout and related components.
ContactBook preview


src/frontend/web_application/src/scenes/ContactBook/components/TagList/index.jsx
```diff
@@ -0,0 +1,83 @@
+import React, { PropTypes } from 'react';
+import { v1 as uuidV1 } from 'uuid';
+import Link from '../../../../components/Link';
+import NavList, { ItemContent } from '../../../../components/NavList';
+
+import './style.scss';
+
+function nbContactsTag(all, tag) {
```
iamdey: I don't understand this function.

I think we can do it without the while instruction (eslint doesn't like it ? ;) ).

At least, can you add the types of the arguments and a description of this function ? E.G

```
/**
 * Count the number of ... by ...
 * @param Array(<string>) all
 * @param <string> tag 
 * @return <number>
 **/

PR #162, opened 2017-02-15
optimize sticky calc and reflow on scroll

iamdey:


src/frontend/web_application/src/layouts/Page/components/Navigation/components/StickyNavBar/index.jsx
```diff
@@ -25,8 +25,11 @@ class StickyNavbarClass extends Component {
 
   handleScroll(ev) {
```
ziir: That function is called way too many times, you'll need to either throttle or debounce it ..
See https://css-tricks.com/debouncing-throttling-explained-examples/ for examples

src/frontend/web_application/src/layouts/Page/components/Navigation/components/StickyNavBar/index.jsx
```diff
@@ -25,8 +25,11 @@ class StickyNavbarClass extends Component {
...
     const scrollSize = ev.srcElement.documentElement.scrollTop || window.scrollY;
```
ziir: (the reflow is here)

ziir:
commits could be squashed and with the MR, could be given a better description, this is not an optimization per say, it's more of a bug fix.

src/frontend/web_application/src/layouts/Page/components/Navigation/components/StickyNavBar/index.jsx
```diff
@@ -1,6 +1,43 @@
 import React, { Component, PropTypes } from 'react';
 import classnames from 'classnames';
 
+const throttle = (func, wait) => {
```
iamdey: I didn't need that kind of library (lodash or whatever) since es6 has almost all required method.

But yep you right, my fork has no unit tests, that's a shame. I may have saw that I can require only this function from lodash.

src/frontend/web_application/package.json
```diff
@@ -63,6 +63,7 @@
     "iron": "^4.0.4",
     "jquery": "^3.1.1",
     "locale": "^0.1.0",
+    "lodash": "^4.17.4",
```
ziir: Can you share a shrinkwrap of the application's dependency tree?

If `lodash` is not already there, you might want to use the standalone lodash throttle package `https://www.npmjs.com/package/lodash.throttle`

ziir:
:+1: Thanks a lot

iamdey:
thank you @ziir, to take time and review it.

PR #168, opened 2017-02-23
[wip]Feature/backend/device patch

gdchamal:
Add PATCH /devices/<device_id> method.

locations (list of network addresses) related to a device are also managed

### DIFF
--- a/src/backend/main/py.main/caliopen_main/user/store/device.py
+++ b/src/backend/main/py.main/caliopen_main/user/store/device.py
@@ -18,8 +18,8 @@ log = logging.getLogger(__name__)
 class DeviceLocation(BaseUserType):
     """Device known location, base on IP."""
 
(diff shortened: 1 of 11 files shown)


src/backend/main/py.main/caliopen_main/user/store/device.py
```diff
@@ -8,16 +8,25 @@
...
 
 log = logging.getLogger(__name__)
 
 
-class Device(BaseModel):
+class DeviceLocation(BaseUserType):
+    """Device known location, base on IP."""
+
+    location_ip = columns.Text(primary_key=True)    # IP address with CIDR
+    location_type = columns.Text()                  # home/work/etc
```
iamdey: if it can be edited by the user I would say it's a `location_name`. What do you think ?

iamdey:
LGTM

PR #169, opened 2017-02-24
refactor BlockList component

iamdey:
+ remove background-color
+ replace ItemContent from ContactBook by specific style rules
+ replace old hardcoded classname m-block-list__item-content in Discussion list by specific rules

No visual changes

---

initial problems with this kind of code:

```
<ItemContent><Link>Foo</Link></ItemContent
```

* we cannot use ItemContent AND Link together, there's no `:active` pseudo class selector, and the `:hover` selector was applied on the parent, it worked but that is not perfect
* ItemContent provides a padding so we wasn't able to put the `:hover` responsibility on the Link or an alternate default background color.


src/frontend/web_application/src/components/BlockList/components/ItemContent/style.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../../../styles/common';
+
+.m-item-content {
+  padding: $co-component__spacing 0;
+  border-bottom: 1px solid $co-color__bg__back;
```
iamdey: may be we should move the border to the `m-block-list__item` 
then remove completely the ItemContent component and let the responsibility of the padding to the developer. 

WDYT @myleneb ?

myleneb: OK I agree to remove ItemContent componnent.

src/frontend/web_application/src/components/BlockList/style.scss
```diff
@@ -10,27 +10,4 @@
   &__item {
     padding: 0;
```
myleneb: is this still needed ?

PR #173, opened 2017-03-01
fix how raw msg requests are handled backend side to conform to new db model.

sapiens-sapide:
fix how raw msg requests are handled backend side to conform to new db model.

### DIFF
--- a/src/backend/main/py.main/caliopen_main/message/core/raw.py
+++ b/src/backend/main/py.main/caliopen_main/message/core/raw.py
@@ -36,8 +36,16 @@ class RawMessage(BaseCore):
                                              data=raw, size=size)
 
     @classmethod
-    def get(cls, raw_msg_id):
-        """Get raw message by raw_msg_id."""
+    def get_for_user(cls, user_id, raw_msg_id):
+        """
+        Get raw message by raw_msg_id, if message belongs to user
+
+        @param: user_id is a string
+        @param: raw_msg_id is a string
+        """
+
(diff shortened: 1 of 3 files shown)


src/backend/main/py.main/caliopen_main/message/core/raw.py
```diff
@@ -54,3 +62,7 @@ class UserRawLookup(BaseUserCore):
 
     _model_class = ModelUserRawLookup
     _pkey_name = 'raw_msg_id'
+
+    @classmethod
+    def message_belongs_to_user(cls, user_id, raw_msg_id):
+        return super(UserRawLookup, cls).object_belongs_to_user(user_id, raw_msg_id)
```
gdchamal: no super needed, inheritance permit to have method in class definition directly.

AKA cls.object_belong_to_user exists

src/backend/main/py.storage/caliopen_storage/core/base.py
```diff
@@ -150,3 +150,11 @@ def create(cls, user, **attrs):
         """Create a core object belong to an user."""
         obj = cls._model_class.create(user_id=user.user_id, **attrs)
         return cls(obj)
+
+    @classmethod
+    def object_belongs_to_user(cls, user_id, object_id):
```
gdchamal: rename this method to belongs_to_user or something more pythonic, and you don't need to have a message_... object_... structure

so first remark on super disappear

sapiens-sapide: changed with commit #9da1b4d below.

PR #177, opened 2017-03-03
implement tags management in settings

iamdey:
Tags in settings.

capture d ecran de 2017-03-03 01 08 22


iamdey:
I'm still not convinced by this UX. But well we probably need to see it in real.

Plus it will bring complexity due to api constraints : Edition is not allowed. We mustn't fake it with delete/recreate.

PR #178, opened 2017-03-03
change backend's REST API listening port from 3141 to 31415.

sapiens-sapide:
Proposal to change backend's REST API listening port from 3141 to 31415, as 3141 is a (old) referenced port instead of 31415 which is free..


iamdey:
Caliopen will be a success only because of this awesome port

+1

(I don't understand why it is not yet taken)

Le ven. 3 mars 2017 13:14, Stanislas SABATIER <notifications@github.com> a
écrit :

PR #180, opened 2017-03-03
Implement discussion API

iamdey:
Some containers are missing in order to load messages in Caliopen, so some mocks are rendered when the API has no data.


src/frontend/web_application/src/store/middlewares/discussions-middleware.js
```diff
@@ -0,0 +1,12 @@
+import { getNextOffset, requestDiscussions, LOAD_MORE_DISCUSSIONS } from '../modules/discussion';
+
+export default store => next => (action) => {
+  const result = next(action);
+
+  if (action.type === LOAD_MORE_DISCUSSIONS) {
+    const offset = getNextOffset(store.getState().discussion);
```
ziir: might want to use a selector here

src/frontend/web_application/src/scenes/DiscussionList/index.js
```diff
@@ -0,0 +1,25 @@
...
+import { connect } from 'react-redux';
+import { withTranslator } from '@gandi/react-translate';
+import { requestDiscussions, loadMoreDiscussions, hasMore } from '../../store/modules/discussion';
+import Presenter from './presenter';
+
+const discussionSelector = state => state.discussion;
+const mapStateToProps = createSelector(
+  [discussionSelector],
+  state => ({
+    discussions: state.discussions.map(discussionId => state.discussionsById[discussionId]),
```
ziir: might want to use a selector here

iamdey: It isn't already one ?

I probably need to refactor all my selectors for correct memoization. It's in my todo list.

PR #184, opened 2017-03-07
Document how we declare first device and how sign API calls

gdchamal:
2 UML diagrams to show how we can declare the first device upon user registration
and how to sign API calls using a device ECDSA private key stored crypted in local storage.

Of course loss of password, make the device not usable directly, we must declare it again, we must think about a recover/reset process to cover this problem


doc/UML/user_and_device_authentication.uml
```diff
@@ -0,0 +1,24 @@
+@startuml
+
+title A known device login on a Caliopen instance
+
+== Authentication ==
+User -> Device: give username and password
+Device -> Device: private_key = AES.decrypt(crypted_key, password)
+Device -> Device: sign_auth = private_key.sign(password)
```
iamdey: when and how you check `sign_auth` on backend side ?

doc/UML/user_and_device_authentication.uml
```diff
@@ -0,0 +1,24 @@
...
+
+Device -> API: POST /authentications (username, password, sign_auth)
+API -> Backend: check_credentials
+Backend -> API: access token
+API -> Backend: check_device_signature(device_id, sign(device_id))
```
iamdey: is this here ?

doc/UML/user_and_device_authentication.uml
```diff
@@ -0,0 +1,24 @@
...
+Backend -> API: ok
+API -> Device: 200
+
+== Call API ==
+Device -> Device: sign_call = sign(access_token)
+Device -> API: GET /discussions with headers Access-Token, X-Device-ID X-Device-Signature
```
iamdey: to be exact access-token is decrypted from cookie on nodejs express middleware

ziir: Yes, in `Authorization` header (https://github.com/CaliOpen/caliopen.web/blob/master/app/routes/api.js#L18)

gdchamal: I set Authorization header in sequence diagram as pointed by @ziir

doc/UML/create_user_and_device.uml
```diff
@@ -0,0 +1,22 @@
+@startuml
+
+title Create a new user and a default device
+
+User -> Device: create a new account (choose username and a password)
+Device -> API: GET /v2/username/isAvailable
+API -> Device: 200 OK
+
+Device -> Device: generate an ECDSA keypair
+Device -> Device: crypt_key = AES.crypt(private_key, password)
```
gdchamal: Yes stored in the local storage, it should be persistent as possible

gdchamal:
merge in this current status, will update later if needed

PR #186, opened 2017-03-07
Implement discussion API

iamdey:


iamdey:
Tests fail after removing mocks, cf. #187

PR #193, opened 2017-03-10
Frontend/ DiscussionList Actions Container

myleneb:
DiscussionList Actions Container css to fit UI


src/frontend/web_application/src/components/Dropdown/style.scss
```diff
@@ -4,7 +4,13 @@
 .m-dropdown {
   @include dropdown-container;
   width: auto;
-
   background-color: $co-color__fg__back;
   box-shadow: $co-shadow;
+
+  &--flat {
```
iamdey: ok the problem comes from predefined styles by foundation. I had the same pbm with ApplicationSwitcher. 

So you only need to use the `Raw` component instead of Dropdown then make your own styles

src/frontend/web_application/src/components/Dropdown/presenter.jsx
```diff
@@ -11,10 +11,19 @@ Raw.propTypes = {
   children: PropTypes.node.isRequired,
 };
 
-const Presenter = ({ className, ...props }) => {
+
+const Presenter = ({ className, flat, ...props }) => {
+  const DropdownClassName = classnames(
+    'm-dropdown',
+    {
+      'm-dropdown--flat': flat,
```
iamdey: no flat modifier, use Raw instead (it is not well named or not enough documentation :(, my bad )

src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/presenter.jsx
```diff
@@ -66,8 +66,13 @@ class DiscussionItemActionsContainer extends Component {
         >
           {__('discussion-item-actions.action.more')}
         </DropdownController>
-        <Dropdown id={this.dropdownId} position="bottom" closeOnClick>
-          <VerticalMenu>
+        <Dropdown
```
iamdey: So:

```
import { Raw } from '../<path>/Dropdown';

// ...
<Raw 
  id={this.dropdownId} 
  position="bottom" 
  closeOnClick 
  className="m-discussion-item-actions-container__dropdown"
>
 //...
```

src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/style.scss
```diff
@@ -1,15 +1,54 @@
...
+    border-left: 1px solid $co-color__primary--lower;
+    line-height: 2 * $co-component__height;
+
+    &:first-child {
+      border-left: 0;
+    }
+  }
+
+  &__menu {
+    margin-top: -1px; // to avoid blank before block (because of Foundation 'Dropdown' JS)
```
iamdey: may be fixed with `Raw` dropdown

src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/style.scss
```diff
@@ -1,15 +1,54 @@
...
+      background: $co-color__primary--low;
+      color: $co-color__fg__text--higher;
+    }
+
+    .s-discussion-list__thread {
+      color: $co-color__primary__text;
+
+      &:hover,
+      &:active {
+        color: $co-color__primary__text;
```
iamdey: useful ?

src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/presenter.jsx
myleneb: mmm I thought the good way to call Raw was via `raw` prop:
```
<Dropdown
  id={this.dropdownId} 
  position="bottom" 
  closeOnClick 
  raw
  className="m-discussion-item-actions-container__dropdown"
>
```
(the other way doesn't seem to work)

but! the only difference I see between Dropdown and Raw is the class `.m-dropdown`, which is responsible for the foundation Dropdown styles, including position and visibility (via `@include dropdown-container;`). 

So imo:
- className `m-dropdown` should alway be applied to Dropdown comp. even the raw one;
- `.m-dropdown` should only contain foundation-dropdown styles (no border, no background, no shadow)
- there should be modifiers for stylized Dropdown (ex: `--dark` for current user menu and application Switcher) 

What do you think of that?

iamdey: > So imo:
>
>  *  className m-dropdown should alway be applied to Dropdown comp. even the raw one;

Agree but don't forget the AppSwitcher component that has very specific styles

> *   .m-dropdown should only contain foundation-dropdown styles (no border, no background, no shadow)

not agree, we should completely drop this mixin and make our own because it adds a lot of things (like border, background) and we should not remove it with override directives ("easy to add, hard to remove"): https://github.com/zurb/foundation-sites/blob/develop/scss/components/_dropdown.scss#L42-L60

>  *  there should be modifiers for stylized Dropdown (ex: --dark for current user menu and application Switcher)

+1

---

Well, we 2 things to deal with in dropdown's stylesheets: 
* the business logic: the positioning (position: absolute, z-index, visibility)
* the styles depending on the context: app switcher, user-menu, discussion-list-more-actions-menu, other-menu

So I propose to refactor the dropdown like this:

* Dropdown only gives business logic
* AppSwitcher extends Dropdown with its special styles
* Add new DropdownMenu component with styles and a standard stylized Button
* the other components will used DropdownMenu and Button

-> If you are ok we can do that in an other PR 
(probably me, because you have enough work with the public site in this sprint)

PR #199, opened 2017-03-15
yaml modelization for new 'message' object and related objects 

sapiens-sapide:
(+ the json representation).
YAML is for ease of read and annotations.

### DIFF
--- a/src/backend/defs/rest-api/objects/MessageV2.yaml
+++ b/src/backend/defs/rest-api/objects/MessageV2.yaml
@@ -1,0 +1,59 @@
+---
+type: object
+properties:
+  attachments:
+    type: array
+    items:
+      "$ref": Attachment.json
+  body: # in future, could be only a preview of body, limited to xxx chars
+    type: string
+  date:
+    type: string
+    format: date-time
+  date_delete:
+    type: string
+    format: date-time
+  date_insert:
+    type: string
+    format: date-time
+  discussion_id:
+    type: string
+  external_references:
+    type: object
+    "$ref": ExternalReferences.json
+  identities: # to which user's local identities the message is linked to
+    type: array
+    items:
+      "$ref": LocalIdentity.json
+  importance_level:
+    type: integer
+    format: int32
+  parent_id: # The id of the Message this message is a reply to. This is primarily for drafts, but the server may support this for received messages as well by looking up the [@!RFC5322] Message-Id referenced in the In-Reply-To header and searching for this message in the user’s mail.
+    type: string
+  is_answered:
+    type: boolean
+  is_draft:
+    type: boolean
+  is_unread:
+    type: boolean
+  message_id:
+    type: string
+  privacy_features:
+    type: object
+    "$ref": PrivacyFeatures.json
+  raw_msg_id:
+    type: string
+  recipients:
+    type: array
+    items:
+      "$ref": Recipient.json
+  subject:
+    type: string
+  tags:
+    type: array
+    items:
+      "$ref": Tag.json
+  type:
+    type: string
+  user_id:
+    type: string
--- a/src/backend/defs/rest-api/objects/Recipient.yaml
+++ b/src/backend/defs/rest-api/objects/Recipient.yaml
@@ -1,0 +1,17 @@
+---
+type: object
+properties:
+  address:
+    type: string
+  contact_id:
+    type: string
+  label: # 'display-name' for email ; 'real name' for IRC
+    type: string
+  protocol: # email, IRC…
+    type: string
+  type: # to, from, cc, bcc…
+    type: string
+required:
+- address
+- type
+additionalProperties: false
--- a/src/backend/defs/rest-api/objects/Attachment.json
+++ b/src/backend/defs/rest-api/objects/Attachment.json
@@ -1,0 +1,18 @@
+{
+  "type": "object",
+  "properties": {
+    "content_type": {
+      "type": "string"
+    },
+    "name": {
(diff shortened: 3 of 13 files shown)


src/backend/defs/rest-api/objects/MessageV2.yaml
```diff
@@ -0,0 +1,59 @@
...
+  importance_level:
+    type: integer
+    format: int32
+  parent_id: # The id of the Message this message is a reply to. This is primarily for drafts, but the server may support this for received messages as well by looking up the [@!RFC5322] Message-Id referenced in the In-Reply-To header and searching for this message in the user’s mail.
+    type: string
+  is_answered:
+    type: boolean
+  is_draft:
+    type: boolean
+  is_unread:
```
iamdey: positive positivism ? what about: `is_read`

src/backend/defs/rest-api/objects/MessageV2.yaml
```diff
@@ -0,0 +1,59 @@
+---
+type: object
+properties:
+  attachments:
+    type: array
+    items:
+      "$ref": Attachment.json
+  body: # in future, could be only a preview of body, limited to xxx chars
```
iamdey: What about a [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern) or [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) for this kind of behavior ?
The `attachments` and `subject` property can be included in those patterns. 

What I mean is a communication is not only simple text. It can be raw email text content, html, markdown, visio call, audio call, encrypted text, and what next ...? 

Message must be flexible at this point.

src/backend/defs/rest-api/objects/MessageV2.yaml
sapiens-sapide: we choose the 'is_unread' flag to be closer to the [jmap](http://jmap.io) specification, and because it is the default state of an incoming message. Futhermore, the 'unread' status is the status that application and user are aware of (for displaying unread count for example). Quoting jmap : « Although this may seem inconsistent at first, it actually makes more sense. The “special” status is when a message is unread (this is what clients are interested in), so like isDraft, isFlagged and isAnswered, we make the special status equal to true. It is also consistent with the need for unread counts in mailbox objects, not read counts, and makes the definition of sorting a message list the same for isFlagged and isUnread when conversations are collapsed. »

src/backend/defs/rest-api/objects/MessageV2.yaml
sapiens-sapide: I don't understand the model you are thinking about for the body field… Could you be more specific ?

gdchamal: I've no idea of the correct design pattern we will use if any, but yes we will have to extract pertinent informations from whole message to get the valid representation for client visualization.

For example inline attachment will be part of the body

iamdey: message_schema

It's more about class design, but the data model must permit it

src/backend/defs/rest-api/objects/MessageV2.yaml
```diff
@@ -0,0 +1,59 @@
+---
+type: object
+properties:
```
iamdey: I don't get where is defined the author of the message, it can be:
* the user (in a specific LocalIdentity)
* a known contact 
* an unknown address

(I understood that a message can still have multiple recipients others than me, which can be known or not)

sapiens-sapide: the author of the message is defined in 3 places : 

* user_id
* recipients : `{type: "from", label: "my display name", protocol: "email", contact_id: "aienraseat", address: "me@mydomain.net"}, {type: "reply-to", …`
* identities: `{display_name: "my display name", identifier: "me@mydomain.net", type: "email_address", user_id: "anierstn"}`

In the UI, the user should choose a "Local Identity" to send the message from. This choice creates a local_identity  and add a recipient with type "from" within the draft.

sapiens-sapide: on va créer une route (très basique pour le moment) `/identities/locals` sur laquelle tu pourras faire un GET et récupérer la "local identity" par défaut de l'utilisateur.

src/backend/defs/rest-api/objects/MessageV2.yaml
gdchamal: may we will not use object oriented language and class design is not a consideration ;)

And yes focus is on data model first

src/backend/defs/rest-api/objects/MessageV2.yaml
```diff
@@ -0,0 +1,59 @@
...
+  is_unread:
+    type: boolean
+  message_id:
+    type: string
+  privacy_features:
+    type: object
+    "$ref": PrivacyFeatures.json
+  raw_msg_id:
+    type: string
+  recipients:
```
iamdey: `participants` !!! :)

src/backend/defs/rest-api/objects/Recipient.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
iamdey: `Participant`

sapiens-sapide: "recipient" changed to "participant" (within next PR)

sapiens-sapide:
closed by PR #201

PR #200, opened 2017-03-16
feature/infra/docker_integration_brokerlmtp_nats_subscriber

ptran32:
This is a docker configuration that can be use for testing the mail delivery stack. 
It is composed of broker lmtp, NATS and python subscriber.

Nginx and SMTP servers are  not configured yet. 

They will be present in a next PR


devtools/docker-compose.yml
```diff
@@ -65,3 +65,30 @@ services:
...
+          - apiv1:apiv1.dev.caliopen.org
+          - cassandra:cassandra.dev.caliopen.org
+          - elasticsearch:es.dev.caliopen.org
+          
+      ports:
+        - "2525:2525"
+        - "9042"
+  # NATS
+  nats:
+      build: ../src/backend/interfaces/NATS
```
gdchamal: may get directly the official nats images instead of building it if there is nothing particular to configure

devtools/docker-compose.yml
```diff
@@ -65,3 +65,30 @@ services:
           - api:api.dev.caliopen.org
       ports:
         - "4000:4000"
+  # Broker
+  broker:
+      build: ../src/backend/protocols/go.smtp
+      links: 
+          - apiv1:apiv1.dev.caliopen.org
```
gdchamal: I think apiv1 is not needed for broker

devtools/docker-compose.yml
```diff
@@ -65,3 +65,30 @@ services:
...
+          - elasticsearch:es.dev.caliopen.org
+          
+      ports:
+        - "2525:2525"
+        - "9042"
+  # NATS
+  nats:
+      build: ../src/backend/interfaces/NATS
+      links:
+          - apiv1:apiv1.dev.caliopen.org
```
gdchamal: link on apiv1 for nats ?

devtools/docker-compose.yml
```diff
@@ -65,3 +65,30 @@ services:
...
+          - api:api.dev.caliopen.org
+          - internal_handler:internal_handler.dev.caliopen.org
+      ports:
+          - "4222:4222"
+  # Internal Message Handler
+  internal_handler:
```
gdchamal: a bit too much generic as naming 'internal_handler'
It does not describe what is it for. I'm thinking about something like 'message_handler', 'message_processor'

devtools/docker-compose.yml
```diff
@@ -65,3 +65,30 @@ services:
...
+      build: ../src/backend/interfaces/NATS/py.client
+      links:
+          - cassandra:cassandra.dev.caliopen.org
+          - elasticsearch:es.dev.caliopen.org
+          - broker:broker.dev.caliopen.org
```
gdchamal: it's strange to have a link to broker instead of one with nats

src/backend/interfaces/NATS/py.client/caliopen.yaml.template
```diff
@@ -0,0 +1,32 @@
+delivery_agent:
```
gdchamal: this file should not be duplicated for docker build purpose.

As we do a complete get of repository in the container image, the /src/backend/configs/caliopen.yaml.template exist and a copy of it in the correct directory should do the job instead of such duplication

PR #201, opened 2017-03-16
New swagger definitions for messages objects & routes [and related]

sapiens-sapide:
This PR also merges PR #198 and PR #199.


src/backend/defs/rest-api/objects/Discussion.json
```diff
@@ -5,10 +5,10 @@
       "type": "integer",
       "format": "int32"
     },
-    "contacts": {
+    "participants": {
       "type": "array",
       "items": {
-        "$ref": "Contact.json"
+        "$ref": "Participant.json"
```
gdchamal: this will break current GET /discussions route
- adapt parameters to this new one
- postpone this update when we will rework this route and related

sapiens-sapide:
Commit #5a8c47b converts all files to yaml and fixes few bugs.
At this stage, we are ready to generate a new swagger.json, but we won't do it until the backend will be ready for the new routes.

src/backend/defs/rest-api/paths/messages.yaml
```diff
@@ -0,0 +1,248 @@
...
+    - name: limit
+      in: query
+      required: false
+      type: integer
+      description: number of messages to return per page
+    - name: offset
+      in: query
+      type: integer
+      required: false
+      description: number of pages to skip from the response
```
iamdey: Is that a way to define an order by ?

sapiens-sapide: yes, we should definitively add more filtering/ordering options on this route, but for now, even the « offset » option does not already work… WIP.

src/backend/defs/rest-api/objects/MessageV2.yaml
```diff
@@ -0,0 +1,73 @@
...
+  user_id:
+    type: string
+required:
+  - date
+  - date_insert
+  - discussion_id
+  - identities
+  - is_answered
+  - is_draft
+  - is_inread
```
iamdey: inread

PR #202, opened 2017-03-17
implement basic message list component with actions on hover

iamdey:
I've added classNames but no styles


myleneb:
As you can see below, responsive UI may have to be improved, especially for MessageActionsContainer:

MessageList responsive

maybe "icon only" buttons for this view?
poke @Lauth0

src/frontend/web_application/src/components/MessageList/components/MessageActionsContainer/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+import './style.scss';
+
+const MessageActions = ({ __, isActive, ...props }) => {
+  if (!isActive) {
+    return (<div {...props} />);
+  }
+
+  return (
+    <div {...props}>
```
iamdey: may be it can simplified:

```
<div className={classnames('m-message-item-actions-container', className)} {...props}>
  ...
```

PR #204, opened 2017-03-21
refactor Dropdown component

iamdey:
* add DropdownMenu
* remove "raw" props
* remove DropdownController component
* add withDropdownControl HoC
* set h offset to 0 (no sapces between controller & dropdown)
* default style for DropdownMenu is primary color background
* UserMenu dropdown uses DropdownMenu

capture d ecran de 2017-03-21 15 20 01
capture d ecran de 2017-03-21 15 19 35


src/frontend/web_application/src/components/DropdownMenu/style.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/common';
+@import '../../styles/object/o-triangle';
+
+.m-dropdown-menu {
+  background-color: $co-color__primary--lower;
+
+  &::before {
```
myleneb: shouldn't triangle be optional?

iamdey: :100: agree but it's not required for now

And in the same way, triangle position should stick with `alignment` property (but not useful too)

PR #206, opened 2017-03-23
[wip]Release/package definition

gdchamal:
Related to https://tree.taiga.io/project/caliopen-caliopen/task/186?kanban-status=815893

The package definition is a simple yaml file


gdchamal:
example of python package resulting:

```
(.venv) tao~/dev/caliopen/CaliOpen/devtools(release/package-definition|…) % ls -ltr ../dist 
total 96
-rw-r--r-- 1 mric mric 53545 23 mars  11:53 caliopen_main-0.0.2-py2.py3-none-any.whl
-rw-r--r-- 1 mric mric 14265 23 mars  11:56 caliopen_storage-0.0.2-py2.py3-none-any.whl
-rw-r--r-- 1 mric mric 23297 23 mars  11:57 caliopen_api-0.0.2-py2.py3-none-any.whl
```

devtools/packages.yaml
```diff
@@ -0,0 +1,59 @@
+---
+# Javascript packages
+web-server:
+  objective: proxy api, provides html & assets (js)
```
iamdey: may be `description` instead

devtools/packages.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - "/dist/windows-phone-app"
+
+# Python packages
+caliopen_storage:
+  objective: Caliopen python storage package
+  release-command: devtools/make_package.sh -t python -s src/backend/main/py.storage -n caliopen_storage
+  dependencies:
+    python: "^2.7"
+  packages:
+  - "/dist/python"
```
iamdey: can it be something like `/dist/caliopen_storage` ?

iamdey: and in this folder we can prevent files overwrite by adding a version e.g:

```
caliopen_storage-0.0.2-py2.py3-none-any.whl
caliopen_storage-0.0.3-py2.py3-none-any.whl
caliopen_storage-0.1.0-py2.py3-none-any.whl
```

devtools/make_package.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+#
+# Script to make a Caliopen package
+#
+
+BASEDIR="`pwd`/.."
+DISTDIR="${BASEDIR}/dist"
+
+
+function make_python_package() {
```
iamdey: I prefer when libs are self package-able

devtools/packages.yaml
gdchamal: yes description sound better

devtools/packages.yaml
gdchamal: Difficult to add the version number in the filename and not well suited a directory per python package.

What about a /dist/<release_number>/<package_name>

devtools/make_package.sh
gdchamal: I don't understand this point

iamdey: I mean, I think this function should be inside of the package source directory, as it can be if we weren't in a mono-repo.

This script (make_package.sh or it could be make_release) should know witch package to build (thanks to package.yaml) but not how to package each part of the app.  
And each source package directory knows how to package itself.

devtools/packages.yaml
iamdey: :+1:

devtools/make_package.sh
gdchamal: I agree with that, but it sound difficult as we must put build packages in a directory outside of the packages source namespace

/src/backend/main/py.main will build a caliopen_main.wheel in /dist directory for example

And in fact build process will be the same for all packages in a given language, that why I think this make_package.sh with a -t <language> option can make things simpler than repeating bulld instruction in all packages.

iamdey: I was thinking about a target directory argument in order to put the dist into where we want (or current working directory if none)
(eg. I want to package a temporary build for whatever ... integration test, quick fix of the front or backend to share with others, build a particular commit without overwriting other stable builts with same version etc.)

devtools/make_packages
```diff
@@ -0,0 +1,130 @@
+#!/bin/env python
+"""Caliopen package creation logic.
+
+Read package definition from a yaml file and build the corresponding
+package depending on language.
+
+cordova:
+  lauguage: javascript
```
iamdey: `language`

devtools/make_packages
```diff
@@ -0,0 +1,130 @@
...
+import logging
+import yaml
+
+log = logging.getLogger(__name__)
+logging.basicConfig()
+log.setLevel(logging.DEBUG)
+
+
+PYTHON_BUILD_CMD = 'python setup.py bdist_wheel --universal --dist-dir {}'
+JS_BUILD_CMD = 'yarn release:web:{}'
```
iamdey: actually this command pattern might not work shortly, I will have something like:

* `yarn run release:web` (does not exist yet. And `release:web:server` and `release:web:browser` build only one package)
* `yarn run release:app:ios`
* `yarn run release:app:android`
* `yarn run release:app:windows`
...

Also all build commands cannot be launched on one machine only: I don't think we can build for android, ios and windows on Linux for example.

Today, `yarn run release` build all available packages in one time, but I have to change it for this previous reason.

And later my goal is to split apps by package or technology in `frontend` directory e.g.:

```
src/frontend/
  + web/
  + cordova/
  + electron/
```

gdchamal:
I discard this PR as I rework most of principle to cover not only package or binary build

PR #213, opened 2017-03-30
Frontend/fixes/reply form

myleneb:
Styles for ReplyForm, and also:
- update font size variables,
- update locales,
- _contrary to what was announced in one commit title, I didn't setup `Dropdown` for protocol selector_


src/frontend/web_application/src/components/ReplyForm/presenter.jsx
```diff
@@ -32,6 +33,8 @@ class ReplyForm extends Component {
     super(props);
     this.state = {
       draftMessage: {},
+      isActive: true,
+      protocol: 'email',
```
iamdey: actually the user don't choose the protocol, it is the contact's available protocols witch determine what protocols will be used. Besides the user should be able to select the identity he wants to use.

Well, we don't have all of those things, so it is just fine like this :).

src/frontend/web_application/src/components/DiscussionDraft/index.jsx
```diff
@@ -13,6 +13,18 @@ TopRow.defaultProps = {
   className: null,
 };
 
+export const BottomRow = ({ className, ...props }) => (
+  <div className={classnames('m-discussion-draft__top-row', className)} {...props} />
```
myleneb: oops

PR #217, opened 2017-03-31
implement messages api

iamdey:
GET filtered by dicsussion_id and PATCH to set is_unread=false


src/frontend/web_application/src/scenes/MessageList/index.js
```diff
@@ -0,0 +1,30 @@
+import { createSelector } from 'reselect';
+import { bindActionCreators, compose } from 'redux';
+import { connect } from 'react-redux';
+import { withTranslator } from '@gandi/react-translate';
+import { requestMessages, updateMessage } from '../../store/modules/message';
+import Presenter from './presenter';
+
+const messagesSelector = state => state.message.messages;
+const messageByIdSelector = state => state.message.messagesById;
```
ziir: Not sure to follow what's happenned here, are the messages duplicated in the state?

iamdey: no, the state has a list of messageIds and an object w/ messages referenced by their id. In this case it is not much useful but it is consistent with discussions, tags etc.

ziir: So, if I get this right:

```js
state.messages.messages = [1, 2, 3, 4];
state.messages.messagesById = {
  1: {
     foo: 'bar',
   },
  2: {
     foo: 'baz',
   },
  3: {
     foo: 'oof',
   },
  4: {
     foo: 'zab',
   },
};
```

Yes?

iamdey: yes

iamdey: it sounds to be a common pattern in redux world

ziir: Oh, OK!

Then, don't you think you'd be better off with something like that ?

```js
state.messages.messages = {
  1: {
     id: 1,
     foo: 'bar',
   },
  2: {
     id: 2,
     foo: 'baz',
   },
  3: {
     id: 3,
     foo: 'oof',
   },
  4: {
     id: 4,
     foo: 'zab',
   },
};

const selectMessages = state => state.messages.messages;
const selectMessagesIds = createSelector([selectMessages], messages => Object.keys(messages));
const selectMessagesArray = createSelector([selectMessages], messages => Object.values(messages));
const selectMessageById = createSelector([selectMessages, (state, props) => props], (messages, props) => messages[props]);

```

iamdey: I'm agree with you, this sounds enough to work in most of my cases and I always be aware of that.

But I prefer to keep this as it is since it is recommended by the [documentation](http://redux.js.org/docs/basics/Reducers.html#note-on-relationships) and since I don't have all of use cases in mind, for example I would like to keep an ordered list in state.

PR #220, opened 2017-04-04
Frontend/feature/Multidimensional PI

myleneb:
This is a first attempt to display PI(s)


src/frontend/web_application/stories/components/MultidimensionalPi.jsx
```diff
@@ -0,0 +1,40 @@
...
+//  { name: 'technical', level: 50 },
+]
+export default () => (<MultidimensionalPi piMax={ piMax } pi={ points } />);
+        `}
+      </Code>
+    </div>
+  );
+};
+
+export default Presenter;
```
iamdey: So this file is not useful anymore since there's info and knobs

src/frontend/web_application/src/components/MultidimensionalPi/components/PiGraph/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+Grid.propTypes = {
+  pi: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+  angle: PropTypes.number.isRequired,
+  axeLength: PropTypes.number.isRequired,
+};
+
+const Polygon = ({ pi, angle, axeLength }) => {
+  const polygonPoints = [];
+  let count = 0;
+  pi.map((p) => {
```
myleneb: "error:  Expected to return a value in this function"
I don't know how to write this function the right way, maybe @iamdey you can help ?

ziir: Looks like you might want to use `Array.prototype.reduce` here:

```js
const polygonPoints = pi.reduce((points, p) => {
  const piLevel = p.level;
  const pointX = axeLength - (piLevel * Math.sin((count * Math.PI) / 180));
  const pointY = axeLength - (piLevel * Math.cos((count * Math.PI) / 180));
  
  return [
    ...points,
    pointX,
    pointY,
  ];
}, []);
```

Though, to follow common practice of `<polygon>` usage, it's probably better to stick with `Array.Prototype.map`:
```js
const polygonPoints = pi.map((p) => {
  const piLevel = p.level;
  const pointX = axeLength - (piLevel * Math.sin((count * Math.PI) / 180));
  const pointY = axeLength - (piLevel * Math.cos((count * Math.PI) / 180));
  
  return `${pointX},${pointY}`
});
```

And for the rendering, in any case:
```jsx
return (
  <polygon
    className="m-pi-graph__shape"
    points={polygonPoints.join(' ')}
  />
);
```

iamdey: Agree with @ziir, I don't know what's the better between maps or reduce in this case, it depends on the attr `point` of polygon.

To answer your specific question, `map` should be used to transform the items of an array so you must return a value:
e.g I have `[1, 2, 3]` and I want `['a', 'b', 'c']`.

What you have in mind I think is a `forEach`: do something for each items of an array:
e.g I have an empty value `foo`, a func `doTheThing` and an array `[1, 2, 3]` then I want foo increments each time the func is true:

```js
let foo = 0;
[1, 2, 3].forEach(item => {
  if (doTheThing(item)) { 
    foo++; 
  }
});
```

It's procedural programing (like `for(...)`). You can also use `reduce`, so you don't need to init `foo`, it's called functional programing.

myleneb: finally opt for `forEach()` instead of `map()` or `reduce()`

src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+  piMax: PropTypes.number.isRequired,
+};
+
+Rating.defaultProps = {
+  className: '',
+};
+
+const Ratings = ({ pi, piMax, averagePi, displayAveragePi }) => (
+  <div className="m-pi-ratings">
+    {displayAveragePi &&
```
myleneb: `displayAveragePi` as props to display or not averagePi

PR #221, opened 2017-04-05
route /api/V2/identities/locals works.

sapiens-sapide:


iamdey:
need swagger spec :)

sapiens-sapide:
you have it within swagger.json, haven't you ?

PR #222, opened 2017-04-06
Remove line with origin because it breaks the docker build

ptran32:
This line is necassary ? @sapiens-sapide 
Because it breaks the docker build with this error:

"cd /go/.cache/govendor/github.com/hashicorp/go-multierror; git reset --hard 7554cd9344cec97297fa6649b055a8c98c2a1e55
fatal: Could not parse object '7554cd9344cec97297fa6649b055a8c98c2a1e55'."

I removed the line "origin" and it works well


gdchamal:
invalided

PR #223, opened 2017-04-06
Release/make release

gdchamal:
This is the main tool to manage a Caliopen release.

Basic usage: 

```
cd devtools
./make_release --conf package.yaml --version 0.0.3 --prepare --build
```

A package definition is needed to prepare them depending on the used language and then build them.

The git index is supposed to be master and sync with origin. A commit will define the release and also a 'release-<version> tag created

Only javascript and python packages managed at this time, and a lot of enhancement


devtools/make_release
```diff
@@ -0,0 +1,350 @@
...
+    if not os.path.isdir(directory):
+        raise Exception('No such package directory {}'.format(directory))
+
+    # Should fail if using package name
+    build_cmd = JS_BUILD_CMD.format(target=package['build_target'])
+    cmd = 'cd {}; {}'.format(directory, build_cmd)
+    res = do_package_command(package, cmd)
+    if not res:
+        return False
+    # Move build distribution
```
gdchamal: I introduce a build_target key for javascript and golang package configuration so we can have different packages build from the same directory (different package name and build_target), but we must have a predicatable way where the build can be found.

I use package['name'] but I'm not sure it will always the case ( web_application/dist/<package_name>)
May add another configuration key (or get it from the package.json ?)

iamdey:
Did not look the whole PR yet, I copy/paste something we might forgot:

>  14:50 dey chamal, je pense à un truc que j'avais zappé dans cette histoire: quel est l'objectif ? avoir une version identique pour tous les packages ? ou faire évoluer chacun d'eux indépendamment
> 14:52 dey ex. ex. concret: si on fait un correctif sur le back et release -> va-t-on republier les apps frontend ?
> 14:52 dey en soit c'est pas grâve, seulement inutile et consomme de la BP
> 14:54 dey autre cas de figure, semver peut perdre de son sens. ex: on remplace completement une fonctionnalité du front (pas de changement sur le back) -> on va publier une majeur ? 

In my previous job, we used [pmsiplan](https://github.com/pmsipilot/pmsiplan) to orchestrate build versions (I'm waiting for some screenshots to make it more understandable). Depending on what we choose, it could be a good tool.

gdchamal:
My plan is hazardous you're right, make all packages being same version number at each release, even if a package doesn't not have any change since last one. Fix of some packages can be done using the last digit of x.x.x version number, but it's not enough.

So instead of updating this package number, we will collect them upon release prepare step, assuming this version number is managed by each package and we must write a release-x.x.x.json file to describe what a release is.

I think we must discuss more about that point.

gdchamal:
```./make_release --version 0.3 build ```

will produce all languages packages in dist/0.3 directory

```./make_release --version 0.3 clean```

will clean previous release build

Issue #224, opened 2017-04-07
Build of lmtp fail


gdchamal:
(.venv) tao~/dev/caliopen/go/src/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp(master|✚3…) % go build github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/cmd/caliopen_lmtpd
# github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/receiver.go:28: undefined: guerrilla.BackendResult
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/receiver.go:28: undefined: guerrilla.Envelope
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/smtp.go:71: undefined: guerrilla.Backend
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/smtp.go:72: undefined: guerrilla.Backend
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/smtp.go:81: unknown field 'AllowedHosts' in struct literal of type guerrilla.ServerConfig
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/smtp.go:93: not enough arguments in call to guerrilla.New
	have (*guerrilla.AppConfig, <T>)
	want (*guerrilla.AppConfig, "github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/vendor/github.com/flashmob/go-guerrilla/backends".Backend, log.Logger)
vendor/github.com/CaliOpen/CaliOpen/src/backend/protocols/go.smtp/smtp.go:93: multiple-value guerrilla.New() in single-value context

using govendor sync and a clean GOPATH

sapiens-sapide:
Sorry, but the revision number for guerrilla dependency is incorrect in vendor.json. I'll fix it.

gdchamal:
fixed

PR #225, opened 2017-04-07
Update govendor and use directly errwrap repository

gdchamal:
Fix some go build issues, but still one with lmtp: https://github.com/CaliOpen/Caliopen/issues/224


sapiens-sapide:
I'll fix the incorrect dependency revision references for errwrap & guerrilla in another commit. But this commit introduces a `govendor update` command that we do not want to execute as it updates dependencies revision to their latest instead of sticking to the revision number written in vendor.json.

gdchamal:
Vendor update permit to go further during my tests but yes must not be done. I'll remove them

Issue #215, opened 2017-03-30
add polyfill for find()


iamdey:
find is not supported by majority of browsers notably for smart devices http://devdocs.io/javascript/global_objects/array/find

babel-polyfill should be added on all build (for now only cordova requires it)

ziir:
I believe this issue to be mis-labeled, to me, it's more about a polyfill and JavaScript cross-browser compatibility strategy.

The `babel-polyfill` way is, IMO, a good one for prototyping without too much hassle, yet it may not suffice for the real world and alternatives exist, each having specific pros and cons.

Perhaps a first draft of the supported browsers and targets would a good start?

iamdey:
the polyfill takes care about:
  * `find`
  * `async/await`
  * eventually `Object.values`
  * ...other eventual es6/7 features

The PR #230 will fix this issue, so I close it now.

@ziir feel free to open a PR if you think about a better solution.

PR #230, opened 2017-04-10
Feature/frontend/connect discussions msg draft

iamdey:
draft-management

DONE : 

* auto-save draft (throttle 5sec)

TODO:

* [x] force save 
* ~~[ ] delete draft~~
* [x] send draft
* [ ] connect real API


iamdey:
Just notice, There's no delete button for the draft on UX actually. Ok with that @Lauth0  ?

(We can use kebab menu eventually)

src/frontend/web_application/src/components/MessageList/components/Message/index.jsx
```diff
@@ -100,7 +108,7 @@ class Message extends Component {
 
   render() {
     const { message, __ } = this.props;
-    const author = message.participants.find(participant => participant.type === 'from');
+    const author = message.participants.find(participant => participant.type === 'From');
```
ziir: This information would probably be better of in `mapStateToProps` via a selector?

src/frontend/web_application/src/components/ReplyForm/presenter.jsx
```diff
@@ -54,21 +49,23 @@ class ReplyForm extends Component {
...
   }
 
   handleChange(ev) {
     const { name, value } = ev.target;
 
     this.setState((prevState) => {
-      const draftMessage = { ...prevState.draftMessage, [name]: value };
-      this.props.onChange({ draftMessage });
+      const draft = { ...prevState.draft, [name]: value };
+      this.props.onChange({ draft });
```
ziir: You might want to do `this.props.onChange({ draft })` in a `setState` callback

src/frontend/web_application/src/scenes/DiscussionList/components/ParticipantsIcon/index.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+import ParticipantIconLetter from '../../../../components/ParticipantIconLetter';
+import Icon from '../../../../components/Icon';
+import './style.scss';
+
+const ParticipantsIcon = ({ discussion }) => {
```
ziir: This component is a good candidate for `componentShouldUpdate()` with re-render rules according to the number of participants

src/frontend/web_application/src/scenes/MessageList/components/ReplyForm/presenter.jsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  componentWillMount() {
+    const { discussionId, draft } = this.props;
+    if (!draft && discussionId) {
+      this.props.requestDraft({ discussionId });
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
```
ziir: I'm not sure about this, won't it result in the draft being fetched twice at ~initial client-side render?

src/frontend/web_application/src/scenes/MessageList/index.js
```diff
@@ -3,23 +3,24 @@ import { bindActionCreators, compose } from 'redux';
...
-  [messagesSelector, messageByIdSelector, discussionIdSelector],
-  (messages, messagesById, discussionId) => ({
+  [messageByIdSelector, discussionIdSelector],
+  (messagesById, discussionId) => ({
     discussionId,
-    messages: messages
-    .map(messageId => messagesById[messageId])
-    .filter(message => message.discussion_id === discussionId),
+    messages: Object.keys(messagesById)
+      .map(messageId => messagesById[messageId])
```
ziir: You could use `Object.values(messagesById)` instead of `Object.keys + map` here

src/frontend/web_application/src/store/modules/message.js
```diff
@@ -87,24 +107,67 @@ export function hasMore(state) {
...
+}
+
+function messageIdsByDiscussionIdReducer(state = {}, action = {}) {
+  if (action.type !== REQUEST_MESSAGES_SUCCESS) {
+    return state;
+  }
+
+  const applyMessageId = (messagesState = [], message) => {
+    const nextState = [...messagesState];
+    if (nextState.indexOf(message.message_id) === -1) {
```
ziir: You should return the current state if there's nothing to update

src/frontend/web_application/src/store/modules/message.js
```diff
@@ -87,24 +107,67 @@ export function hasMore(state) {
...
+  };
+
+  return action.payload.data.messages.reduce((acc, message) => {
+    const { discussion_id: discussionId } = message;
+
+    return {
+      ...acc,
+      [discussionId]: applyMessageId(acc[discussionId], message),
+    };
+  }, { ...state });
```
ziir: Unecessary spread operator

src/frontend/web_application/webpack/webpack.config.browser.js
```diff
@@ -10,6 +10,7 @@ const isProd = process.env.NODE_ENV === 'production';
 
 let config = Object.assign(baseConfig.getBase('browser'), {
   entry: [
+    'babel-polyfill',
```
ziir: So, `babel-polyfill` is the polyfill direction the project is taking?

https://github.com/CaliOpen/Caliopen/issues/215

src/frontend/web_application/yarn.lock
```diff
@@ -3272,7 +3272,7 @@ es6-map@^0.1.3:
     es6-symbol "~3.1.0"
     event-emitter "~0.3.4"
 
-es6-promise@^3.0.2:
+es6-promise@^3.0.2, es6-promise@^3.2.1:
```
ziir: 2 distinct versions of `es6-promise` ?

src/frontend/web_application/yarn.lock
```diff
@@ -7747,7 +7747,7 @@ range-parser@^1.0.3, range-parser@~1.2.0:
   version "1.2.0"
   resolved "https://registry.yarnpkg.com/range-parser/-/range-parser-1.2.0.tgz#f49be6b487894ddc40dcc94a322f611092e00d5e"
 
-raw-body@~2.2.0:
+raw-body@^2.1.7, raw-body@~2.2.0:
```
ziir: 2 distinct versions of `raw-body` ?

src/frontend/web_application/src/components/MessageList/components/Message/index.jsx
iamdey: It's not a connected component. But yes we can eventually memoized the result using reselect. 
For now I think it is micro-optim. 
This kind of selector is almost everywhere, I will do an optimization check after this PR and may be apply memoization in that case if needed.

src/frontend/web_application/src/components/ReplyForm/presenter.jsx
iamdey: it actually is, isn't it ?

src/frontend/web_application/src/scenes/DiscussionList/components/ParticipantsIcon/index.jsx
iamdey: or extends a `PureComponent`. WDYT?

src/frontend/web_application/src/scenes/MessageList/components/ReplyForm/presenter.jsx
iamdey: Actually `requestDraft` doesn't fetch drafts since there's no API for drafts: draft are common message with `is_draft: true`; 
the `requestDraft` will filter the `message` state and initialize a draft in the state (a new one if there is no existing draft message).

src/frontend/web_application/src/scenes/MessageList/index.js
iamdey: Oh! 
.. I just saw http://devdocs.io/javascript/global_objects/object/values, it still looks a bit buggy
so no, I will leave it as it is.

src/frontend/web_application/src/store/modules/message.js
iamdey: yes, is that a pbm ?

src/frontend/web_application/src/store/modules/message.js
iamdey: fixed

src/frontend/web_application/src/store/modules/message.js
iamdey: fixed

src/frontend/web_application/src/components/ReplyForm/presenter.jsx
ziir: I mean, like this:

```js
this.setState(updateState, () => { this.props.onChange({ draft: this.state.draft }); });

src/frontend/web_application/src/scenes/DiscussionList/components/ParticipantsIcon/index.jsx
ziir: `PureComponent` won't be of much help here, the contrary in fact, because it's "magic" and shallow-compares `props` and `nextProps`, not just `discussion.participants.length` which is the trigger here

src/frontend/web_application/src/scenes/MessageList/components/ReplyForm/presenter.jsx
ziir: I see, then it's ok I guess

src/frontend/web_application/webpack/webpack.config.browser.js
iamdey: yes. 
I know, easy to add & hard to remove. 
Right now, I prefer spend time on features than fixing every browser (that will be done while beta-phase).

src/frontend/web_application/src/scenes/MessageList/index.js
ziir: I don't know what you mean by "buggy", though ES2018 Draft, it's already available in core-js/babel

src/frontend/web_application/src/store/modules/message.js
ziir: Unnecessarily creating a new object, and merging the content of another into it, adding junk code after babel-transpile ...

It's absolutely safe to just use `state` here

src/frontend/web_application/webpack/webpack.config.browser.js
ziir: Fine by me, though it could help it being documented somewhere don't you think?

src/frontend/web_application/src/components/MessageList/components/Message/index.jsx
ziir: Ok

PR #232, opened 2017-04-11
Permit to create a user for POC having a domain different thant calio…

gdchamal:
…pen instance default_domain


sapiens-sapide:
Need to manage non default domains within go.smtp & postfix after this merge.

gdchamal:
I may revert this change.

What I want to do for POC data loading will not work, have to define a better and more robust process

PR #235, opened 2017-04-12
Refactor the way nats message string are build in caliopen_nats

sapiens-sapide:
Add a more 'pythonic' way of building the json strings for nats messages.


gdchamal:
LGTM

PR #238, opened 2017-04-13
Move parsing logic into a dedicated namespace caliopen_mail.parsers

gdchamal:
mail parsing was previously into caliopen_mail.message.format namespace, we mut group all parsing logic at the same place (vcard will come)

So a caliopen_main.parsers namespace is created to permit such grouping


sapiens-sapide:
ok

PR #240, opened 2017-04-14
import vcard with several vcard on a file

lisonferez:
Add several NewContact at a user with several vcard in the same file


gdchamal:
:+1:

PR #241, opened 2017-04-14
Link correctly for docker usage of API

gdchamal:
Api will start correctly in a docker container with this configuration

### DIFF
--- a/src/backend/interfaces/REST/go.server/caliopen-api.docker.yaml
+++ b/src/backend/interfaces/REST/go.server/caliopen-api.docker.yaml
@@ -21,4 +21,4 @@ ProxyConfig:
   port: 31415
   routes:
   - /api/v2/: 127.0.0.1:6544
-  - /: 127.0.0.1:6543         # route "/" catch all other requests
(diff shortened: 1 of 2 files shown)


src/backend/interfaces/REST/go.server/caliopen-api.docker.yaml
```diff
@@ -5,15 +5,15 @@ ApiConfig:
...
-    url: nats://localhost:4222
+    url: nats://nats.dev.caliopen.org:4222
     outSMTP_topic: outboundSMTP # topic's name to post "send" draft order
   swaggerSpec: ../../../../../../../doc/api/swagger.json #absolute path or relative path to go.server bin
   RedisConfig:
-    host: localhost:6379
+    host: redis.dev.caliopen.org:6379
     password: ""      #no password set
     db: 0             #use default db
 ProxyConfig:
```
iamdey: I've also noticed: `- /: apiv1.dev.caliopen.org:6543`
wdyt ?

gdchamal: yes you're right that's may work better

PR #245, opened 2017-04-17
Fix/make delivery great again

gdchamal:
Various fixes to make delivery using broker + message_handler works in a docker container


gdchamal:
Bad branch base not on master :/

PR #247, opened 2017-04-18
Feature/backend/remove counter

gdchamal:
This feature remove usage of cassandra counters, a feature that will disappear anyway in the future (and have many corner cases for consistency guaranty).

This permit to replace cassandra backend by Scylladb and take less memory to run the development stack (512MB instead of 2GB for cassandra).

For the moment 2 properties on discussion are not computed, unread_count and attachment_count, they were managed using counters, it will be replace using index compute capabilities


iamdey:
:tada:

PR #251, opened 2017-04-19
fix broker hostname for docker environment in CLI

### LINKED ISSUE #243: config for CLI
unable to change the hostname for the broker. so the import emails command might not work within docker:

https://github.com/CaliOpen/Caliopen/blob/master/src/backend/tools/py.CLI/caliopen_cli/commands/import_email.py#L84

also the link between cli and broker is missing

iamdey:
fix #243 

PS: totally blind fix since, Dockerfile cannot add the fix to its workspace


sapiens-sapide:
Tested & approved.

PR #254, opened 2017-04-20
Add 'Bearer' auth-scheme to api/v2

sapiens-sapide:


src/backend/interfaces/REST/go.server/middlewares/authentication.go
```diff
@@ -25,9 +28,14 @@ func BasicAuthFromCache(cache *redis.Client, realm string) gin.HandlerFunc {
 		// Get provided auth headers
 		var user_id, access_token string
 		var ok bool
-		if user_id, access_token, ok = c.Request.BasicAuth(); !ok {
-			kickUnauthorizedRequest(c, realm)
+
+		//Try auth-scheme 'Bearer' then 'Basic'
+		if user_id, access_token, ok = BearerAuth(c.Request); !ok {
+			if user_id, access_token, ok = c.Request.BasicAuth(); !ok {
```
iamdey: The question is: should we allow multiple ways to authenticate on the backend ?
So this must be documented.

IMO, Bearer does the job and should be the only way (api/v1& api/v2).
If Basic is just a convenient way to authenticate for tests purpose, this should not be available on production code: We may provide an other tool to generate a Bearer (the actual express middleware may be)

sapiens-sapide: Yes, I kept "Basic" auth-scheme for testing purpose because it is the 'official' token for Basic Authentication ((see RFC7616 and RFC7617). All testing tools use this token.

iamdey: so we must use `Basic` instead of `Bearer`. WDYT @gdchamal ?

(I've no idea why we start with "Bearer")

gdchamal: both Bearer and Basic authentication schema must be acccepted by API.

Basic is cool for testing purposes, Bearer is what we must use for real condition (so Front use Bearer as we discussed previously)

PR #257, opened 2017-04-21
Feature/backend/mail qualifier

gdchamal:
Rework around UserMessageQualifier to create the caliopen message resource and associate it to a discussion or create a new one.

There is no more, for the moment, json representation of message with the raw data.

The CLI import email command does not need anymore the whole stack to be up.

### DIFF
--- a/src/backend/brokers/go.emails/email.go
+++ b/src/backend/brokers/go.emails/email.go
@@ -7,7 +7,6 @@ package email_broker
 
 import (
 	"bytes"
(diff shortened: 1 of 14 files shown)


sapiens-sapide:
Need to change order send by brokers/go.emails/inbound.go from `process_message` to `process_email_raw`

gdchamal:
Update nats message to be process_raw between go smtp and python nats handler with af05224

PR #259, opened 2017-04-24
Refactor/frontend/contact book

myleneb:
Connect ContactBook
- add contacts in api-mock
- create reducer, middleware and update
configue-store
- remove sortView function and form
- replace Taglist Links with Buttons
- clean some CSS


src/frontend/web_application/src/scenes/ContactBook/components/TagList/index.jsx
```diff
@@ -20,22 +20,33 @@ function nbContactsbyTag(list, tag) {
...
-    </Button>
-  </ItemContent>
-);
+const TagItem = ({ title, link, onTagClick, nbContacts, active, all, className }) => {
+  const tagClassName = classnames(
+    className,
+    'm-tag-list__tag',
+    {
+      'm-tag-list__tag--active': active,
+      'm-tag-list__tag--all': all,
```
iamdey: I don't understand this modifier: `all`

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -1,95 +1,80 @@
...
+      sortDir: defaultSortDir,
+      sortView: defaultSortView,
     };
+    this.loadMore = this.loadMore.bind(this);
   }
 
   componentDidMount() {
-    this.loadTags();
+    this.props.requestContacts();
     setTimeout(() => {
```
iamdey: no more `setTimeOut` :)
It was used to fake a request.

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -1,95 +1,80 @@
...
-      this.loadContacts();
+      this.loadTags();
```
iamdey: ok, here is the problem: `loadTags` does not fetch anything, it uses contacts to calc all tags.

What I think: 
* first rename `loadTags` into `initializeTags` (or calcT.. or extractTa... or ...), it will sounds more accurate

The first time you try to calc tags, there is no contacts since you just fetch its (fortunately, the request arrives within 2ms so you were able to extract the tags, but it won't work if the response take longer).

* Then, you just need to move this call in the [`
componentWillReceiveProps`](https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops): every time `props.contacts` change, it will recalc tags

src/frontend/web_application/src/store/modules/contact.js
```diff
@@ -0,0 +1,112 @@
...
+export const REQUEST_CONTACTS = 'co/contact/REQUEST_CONTACTS';
+export const REQUEST_CONTACTS_SUCCESS = 'co/contact/REQUEST_CONTACTS_SUCCESS';
+export const REQUEST_CONTACTS_FAIL = 'co/contact/REQUEST_CONTACTS_FAIL';
+export const INVALIDATE_CONTACTS = 'co/contact/INVALIDATE_CONTACTS';
+export const LOAD_MORE_CONTACTS = 'co/contact/LOAD_MORE_CONTACTS';
+export const REQUEST_CONTACT = 'co/contact/REQUEST_CONTACT';
+export const REMOVE_CONTACT = 'co/contact/REMOVE_CONTACT';
+
+export function requestContacts(params = {}) {
+  const { offset = 0, limit = 20 } = params;
```
iamdey: I think limit can be `1000` in this case (pagination is very weird when you search for a contact)

PR #260, opened 2017-04-24
Fix Dropdown prop `position` propType

myleneb:
this add `left` to oneOf[..] `position` propType for DropDown (on ReplyForm and MessageList, it opens on left)


myleneb:
mmm well actually it doesn't impact foundation dropdown script, it's only to avoid propTypes erros. I spent a few time on it but I have to admit, there's still some magic in here. `left` position property works well for dropdowns in MessageList but not in storybook :/ my conclustion is the same we have last time, we should have our own dropdown component not related to foundation.

PR #263, opened 2017-04-26
Fix Taglist loading on page refresh

myleneb:
`TagList` in `ContactBook` component didn't load tags properly on page refresh, this fixes it.

### DIFF
--- a/src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
+++ b/src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
@@ -11,6 +11,15 @@ import './style.scss';
 const defaultSortDir = 'ASC';
 const defaultSortView = 'title';
 
+function generateStateFromProps(props) {
+  const { contacts } = props;
+  const tags = [];
+  contacts.map(contact => contact.tags.map(tag => tags.push(tag)));
+
+  return { tags };
+}
+
+
 function getOrderedContacts(contactList, sortView, sortDir) {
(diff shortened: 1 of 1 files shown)


src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -11,6 +11,15 @@ import './style.scss';
 const defaultSortDir = 'ASC';
 const defaultSortView = 'title';
 
+function generateStateFromProps(props) {
+  const { contacts } = props;
+  const tags = [];
+  contacts.map(contact => contact.tags.map(tag => tags.push(tag)));
```
ziir: If you wanted to make it less imperative, and embrace a little more the `.map` API style:

```js
const tags = [].concat(...contacts.map(contact => contact.tags));
```
If you wish to also make it an array of unique values:

```js
const tags =[...new Set([].concat(...contacts.map(contact => contact.tags)))];
```

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -61,26 +70,21 @@ class ContactBook extends Component {
     this.loadMore = this.loadMore.bind(this);
   }
 
+  componentWillMount() {
+    this.setState(prevState => generateStateFromProps(this.props, prevState));
```
ziir: Might want to set this in the constructor instead

PR #264, opened 2017-04-26
Feature/backend/api/import vcard

lisonferez:
New route in API Rest: POST /import.
This route make it possible to import a file of vcards and insert new contact at a user with vcards.


gdchamal:
I think a rebase from master is missing or your branch doesn't start from master but your previous one (for import vcard documentation) as we see commit not related to this API feature

gdchamal:
Also the modified swagger.json file containing your new route is missing in this PR

gdchamal:
:+1:

src/backend/defs/rest-api/paths/import.yaml
```diff
@@ -0,0 +1,11 @@
...
+  post:
+    consumes:
+    - text/vcard
+    - text/vcf
+    - text/x-vcard
+    responses:
+      '200':
+        description: Import successful
+      '400':
+        description: Syntax error
```
sapiens-sapide: Miss : 
 * "security" directive
 * "produces" directive
 * "description" & "schema" for responses.

I suggest the route to be `/v1/contacts/import` or  `/v1/import/contacts`

gdchamal: Different content-type permit to import differents format.
I think a generic route is not so bad

PR #265, opened 2017-04-26
Feature/backend/go vendoring

sapiens-sapide:
Change the place of Go's vendor directory to enable a global Go dependencies management, at repos' level.


src/backend/configs/caliopen-go-lmtp_dev.yaml
```diff
@@ -24,18 +24,18 @@ AppConfig:
 ## LDA (Email broker) config ##
 LDAConfig:
   broker_type: smtp # types are : smtp, imap, mailboxe, etc.
-  nats_url: nats://localhost:4222
+  nats_url: nats://nats.dev.caliopen.org:4222
```
iamdey: this config shouldn't be restricted to a docker environment ?

I mean this is used for both native & vagrant env, right ?

src/backend/Dockerfile.go-api
```diff
@@ -0,0 +1,15 @@
+FROM golang
+
+RUN go get -u github.com/kardianos/govendor
+RUN go install github.com/kardianos/govendor
```
sapiens-sapide: go install should not be necessary, go get already do it.

src/backend/Dockerfile.go-api
```diff
@@ -0,0 +1,15 @@
...
+
+ADD . /go/src/github.com/CaliOpen/CaliOpen/src/backend
+WORKDIR /go/src/github.com/CaliOpen/CaliOpen/src/backend
+
+RUN govendor sync -v
```
sapiens-sapide: to be sure to use the govendor bin inside the docker I'll suggest `$GOPATH/bin/govendor`

src/backend/Dockerfile.go-lmtp
```diff
@@ -0,0 +1,15 @@
+FROM golang
+
+RUN go get -u github.com/kardianos/govendor
+RUN go install github.com/kardianos/govendor
+
+ADD . /go/src/github.com/CaliOpen/CaliOpen/src/backend
+WORKDIR /go/src/github.com/CaliOpen/CaliOpen/src/backend
+
+RUN govendor sync -v
```
sapiens-sapide: `govendor sync` => `$GOPATH/bin/govendor sync`

src/backend/configs/caliopen-go-lmtp_dev.yaml
sapiens-sapide: yes

PR #268, opened 2017-04-27
Build/frontend/copy local code

iamdey:


devtools/docker-compose.yml
```diff
@@ -70,8 +70,7 @@ services:
 
   # Caliopen frontend
   frontend:
-      build:
-        context: ../src/frontend/
```
gdchamal: context syntax is recommanded for compose v2 format

src/frontend/web_application/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+# Author: Caliopen
+# Date: 2016-01-05
+
+FROM node:6
+MAINTAINER Caliopen
+
+ADD . /srv/caliopen/frontend/
+WORKDIR /srv/caliopen/frontend/
+RUN yarn install
+RUN yarn run release
```
gdchamal: I think yarn release is not OK as it will use NODE_ENV=production value

src/frontend/web_application/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+EXPOSE 4000
+
+CMD ["yarn", "run", "start:prod"]
```
gdchamal: start instead of start:prod to permit hot reload mechanism

src/frontend/web_application/Dockerfile
iamdey: I would like to go for hot-reload in an other PR, it has more side effects that just changing an env or a command.

I know this PR is stable and does what is important right now: build the current code and start in an optimized  way.

I have no time to polish this PR and I'm not sure it is efficient right now:
* if you are a python/golang hacker: just change the code rebuild and go: `docker-compose build frontend && docker-compose stop frontend  && docker-compose up -d frontend`
* if you are a frontend dev: install, node, yarn, launch the app  `yarn start` then dev (with hot-reload)

WDYT ?

devtools/docker-compose.yml
iamdey: I don't see where it is recommended, but ok

src/frontend/web_application/Dockerfile
sapiens-sapide: As discussed with gdchamal, let's go with Docker containers that are 'production' ready.  
If a developer needs to put his hands in a specific code part, it's better for him to not run the container but run the code barely on its computer. For example, a frontend dev. could launch `redis cassandra elasticsearch`, then `api`, but not `frontend`. And build&run the frontend directly.

PR #269, opened 2017-04-27
Remove py.smtp package as we do not need it anymore.

sapiens-sapide:
Also remove directives in Dockerfile, setup.py, etc. to not build this package anymore.


gdchamal:
:+1:

PR #270, opened 2017-04-28
Fix/frontend/contactbook

myleneb:
bunch of small fixes on ContactBook component:
- `sortView` related functions works (again)
- fix `tags` state in ContactBook
- improve CSS for responsive view (also small fixes on Page layout)


src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -62,17 +59,14 @@ class ContactBook extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      tags: [],
+      tags: [].concat(generateStateFromProps.bind(props)),
```
myleneb: @ziir is it the right way to set tags in constructor? (related to your comment in last PR)

src/frontend/web_application/src/scenes/ContactBook/components/TagList/index.jsx
```diff
@@ -63,28 +63,26 @@ const TagList = ({ tags, onTagClick, nbContactsAll, activeTag }) => {
   const tagList = Array.from(new Set(list));
 
   return (
-    <div>
-      <NavList className="m-tag-list" dir="vertical">
+    <NavList className="m-tag-list" dir="vertical">
+      <TagItem
+        title="All contacts"
```
iamdey: may be add a `__('...')` to make sure we don't forget to translate it

src/frontend/web_application/src/scenes/ContactBook/components/TagList/index.jsx
```diff
@@ -63,28 +63,26 @@ const TagList = ({ tags, onTagClick, nbContactsAll, activeTag }) => {
...
+        link=""
+        nbContacts={nbContactsAll}
+        key={uuidV1()}
+        onTagClick={onTagClick}
+        active={activeTag === '' && true}
```
iamdey: `& true` is useless

src/frontend/web_application/src/scenes/ContactBook/components/TagList/index.jsx
```diff
@@ -63,28 +63,26 @@ const TagList = ({ tags, onTagClick, nbContactsAll, activeTag }) => {
...
-          title="All contacts"
-          link=""
-          nbContacts={nbContactsAll}
+          title={tag}
+          link={tag}
+          nbContacts={nbContactsbyTag(list, tag)}
           key={uuidV1()}
           onTagClick={onTagClick}
-          active={activeTag === '' && true}
+          active={tag === activeTag && true}
```
iamdey: `&& true`

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -8,32 +8,22 @@ import Button from '../../components/Button';
...
 
-  if (sortDir === 'ASC') {
-    sortContacts.sort((a, b) =>
-      a[sortView].localeCompare(b[sortView]));
-  }
-  if (sortDir === 'DESC') {
-    sortContacts.sort((a, b) =>
-      b[sortView].localeCompare(a[sortView]));
-  }
+    if (sortDir === 'ASC') { return first.localeCompare(second); }
```
iamdey: maybe switch case with default on ADD ?

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
iamdey: nope :)

you might not have contacts at this time.
Also I think we should not define calculated values into a constructor: just define the structure of the state, in other word: always use setState.

In our convention, `generateStateFromProps` is used in componentWillMount (when props are ready) and `componentWillReceiveProps` (c.f ReplyForm Component)

ziir: @iamdey nope

https://facebook.github.io/react/docs/react-component.html#constructor
https://facebook.github.io/react/docs/react-component.html#componentwillmount

src/frontend/web_application/src/scenes/ContactBook/components/TagList/index.jsx
```diff
@@ -63,28 +63,26 @@ const TagList = ({ tags, onTagClick, nbContactsAll, activeTag }) => {
...
 
   return (
-    <div>
-      <NavList className="m-tag-list" dir="vertical">
+    <NavList className="m-tag-list" dir="vertical">
+      <TagItem
+        title="All contacts"
+        link=""
+        nbContacts={nbContactsAll}
+        key={uuidV1()}
```
iamdey: it's not in a `map`, key is not useful here and can be removed (plus it will trigger a useless re-render of this tagitem)

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
iamdey: ok right, it can works in constructor but I still prefer the `componentWillMount` way for the readability since it is a common pattern that doesn't break things.

src/frontend/web_application/src/scenes/ContactBook/components/ContactList/index.jsx
```diff
@@ -65,10 +76,10 @@ const ContactList = ({ contacts, sortView }) => {
 
 ContactList.propTypes = {
   contacts: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
-  sortView: PropTypes.string,
+  sortView: PropTypes.oneOf([SORT_VIEW_GIVEN_NAME, SORT_VIEW_FAMILY_NAME]).isRequired,
```
myleneb: this give an error:
Failed prop type: Invalid prop `sortView` of value `given_name` supplied to `ContactList`, expected one of [null,null].
despite thoses variables are imported and used in function... what's the problem?

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
```diff
@@ -98,20 +88,22 @@ class ContactBook extends Component {
...
+    const handleSortViewChange = (event) => {
       this.setState({
         sortView: event.target.value,
       });
-    };*/
+    };
 
     const { contacts, isFetching, hasMore, __ } = this.props;
 
+    const tags = [].concat(...contacts.map(contact => contact.tags));
```
myleneb: @iamdey @ziir finally, don't need tags in state, so I just passed it in render().

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
ziir: @iamdey Please bear with me for saying I believe this is a mistake.

src/frontend/web_application/src/layouts/Page/components/Header/style.scss
```diff
@@ -7,7 +7,6 @@ $l-header__margin: ($co-margin / 2) !default;
 
 .l-header {
   height: $l-header__height;
-  margin-bottom: $l-header__margin;
```
myleneb: yes sticky navbar is ok but maybe I miss something?

src/frontend/web_application/src/scenes/ContactBook/components/ContactList/index.jsx
iamdey: the variables are not well imported, but I don't get why

src/frontend/web_application/src/scenes/ContactBook/components/ContactList/index.jsx
```diff
@@ -1,11 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { v1 as uuidV1 } from 'uuid';
-import BlockList from '../../../../components/BlockList';
 import Title from '../../../../components/Title';
 import ContactItem from './components/ContactItem';
+
 import './style.scss';
 
+import { SORT_VIEW_FAMILY_NAME, SORT_VIEW_GIVEN_NAME, SORT_VIEW_TITLE, DEFAULT_SORT_VIEW } from '../../../ContactBook/presenter';
```
iamdey: this line should be before `import './styles..`

iamdey: may be the pbm comes from an invalid path here

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
iamdey: ok let's include @themouette that introduced me this practice. It is a good to know pro/cons about that.

@ziir you declare using `generateStateFromProps` in `componentWillReceiveProps` is a mistake. Can you explain why ?

src/frontend/web_application/src/scenes/ContactBook/components/ContactList/index.jsx
myleneb: so this is not from invalid import but from oneOf prop type:
`This does not work for values that are imported from external files`
https://github.com/reactjs/react-docgen/commit/d3dbb2e0214c8d5e84089c25a970479186e95383

iamdey: ok may be this proptype can be omitted since it is a reference to child component ... but well, old plain string will do the job

myleneb: https://github.com/CaliOpen/Caliopen/pull/270/commits/f49e312bdac99a82d04358ccd694fd3b5c5a72f5 replaces `oneOf` with `string`

src/frontend/web_application/src/scenes/ContactBook/presenter.jsx
iamdey: :+1:

src/frontend/web_application/src/layouts/Page/components/Header/style.scss
iamdey: ok if it looks ok :)

PR #304, opened 2017-05-10
Feature/frontend/menu bar

myleneb:
This create a component MenuBar to keep consistency for top menu bars on different pages (Setting, MessageList, ContactBook)

This also refactors `<Button>` with new props:
- shape: PropTypes.oneOf(['plain', 'hollow']),
- icon: PropTypes.string,
- display: PropTypes.oneOf(['inline', 'expanded']),
- color: PropTypes.oneOf(['success', 'alert', 'secondary', 'active']),
- responsive: PropTypes.oneOf(['icon-only', 'text-only'])
- accessKey: PropTypes.string

### DIFF
--- a/src/frontend/web_application/src/components/ContactDetails/index.jsx
+++ b/src/frontend/web_application/src/components/ContactDetails/index.jsx
@@ -71,13 +71,15 @@ class ContactDetails extends Component {
 
   renderSubtitleActions() {
     const { __ } = this.props;
-    const activeButtonProp = this.state.editMode && { color: 'active' };
+    const activeButtonProp = this.state.editMode ? { color: 'active' } : {};
 
     return (
       <Button
         className="pull-right"
         {...activeButtonProp}
-        onClick={this.handleSwitchEditMode} icon="edit">
+        onClick={this.handleSwitchEditMode}
+        icon="edit"
+      >
         <span className="show-for-sr">{__('contact.action.edit_contact_details')}</span>
       </Button>
     );
--- a/src/frontend/web_application/src/components/ContactDetails/index.jsx
+++ b/src/frontend/web_application/src/components/ContactDetails/index.jsx
@@ -3,7 +3,6 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Subtitle from '../Subtitle';
 import Button from '../Button';
-import Icon from '../Icon';
 import TextList, { ItemContent } from '../TextList';
 import AddressDetails from './components/AddressDetails';
 import EmailDetails from './components/EmailDetails';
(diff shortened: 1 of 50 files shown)


src/frontend/web_application/src/components/ContactDetails/index.jsx
```diff
@@ -72,13 +71,13 @@ class ContactDetails extends Component {
 
   renderSubtitleActions() {
     const { __ } = this.props;
+    const activeButtonProp = this.state.editMode && { color: 'active' };
```
iamdey: no warning when it's not in editMode ? 
when `activeButtonProp` is filled, the `{...activeButtonProp}` will be changed to `color="active"` on the jsx component. In the other case I don't know exactly what it will be ... may be `undefined="true"` or something else.

---

A ternary operator should be better, I'm pretty sure it will not add strange things:
```
const activeButtonProp = this.state.editMode ? { color: 'active' }) : {};
```

src/frontend/web_application/src/components/ContactDetails/index.jsx
```diff
@@ -72,13 +71,13 @@ class ContactDetails extends Component {
...
+
     return (
       <Button
         className="pull-right"
-        active={this.state.editMode}
-        onClick={this.handleSwitchEditMode}
-      >
-        <Icon type="edit" />
+        {...activeButtonProp}
+        onClick={this.handleSwitchEditMode} icon="edit">
```
iamdey: line breaks after each attribute please

src/frontend/web_application/src/components/MessageList/components/Message/index.jsx
```diff
@@ -1,188 +1,18 @@
...
-          }
-
-        </div>
-      </div>
-    );
-  }
-}
-
-export default Message;
+import { createSelector } from 'reselect';
```
iamdey: ok be carefull, I suppose it is the source of the conflict.

The final file looks good here

src/frontend/web_application/src/scenes/Account/components/AccountOpenPGPKeys/components/AccountOpenPGPKeyForm/index.jsx
```diff
@@ -154,6 +154,8 @@ class AccountOpenPGPKeyForm extends Component {
     // XXX: note on importForm, errors comes from props and form items from this.state since form is
     // not real time saved in redux store
     const { __, isLoading, className, importForm, children } = this.props;
+    const generateHollowProp = this.state.formType === FORM_TYPE_GENERATE && { display: 'hollow' };
+    const rawHollowProp = this.state.formType === FORM_TYPE_RAW && { display: 'hollow' };
```
iamdey: I think it is better to use ternary operator here : ` const myvar = condition ? { foo: 'bar' } : {};`

src/frontend/web_application/src/scenes/Account/components/AccountOpenPGPKeys/presenter.jsx
```diff
@@ -77,14 +77,16 @@ class AccountOpenPGPKeys extends Component {
       onGenerateKey,
     } = this.props;
 
+    const activeButtonProp = this.state.editMode && { color: 'active' };
```
iamdey: same

src/frontend/web_application/src/styles/object/o-clickable.scss
```diff
@@ -57,7 +57,7 @@
 
 @mixin o-clickable--expanded {
   display: block;
-  line-height: inherit;
+  // line-height: inherit;
```
iamdey: yes it is a bit weird but useful when a button has not a regular height (callback buttons for instance).
This permit to vertically align text within large buttons

So ok to remove if you have a better solution

PR #307, opened 2017-05-11
Feature/backend/context reply handling

sapiens-sapide:
This PR modifies how API handles new messages and replies within a discussion.  

From now on, when POSTing a reply with a `discussion_id` property, `subject` and `participants` properties are automatically computed by backend. Frontend should not send those properties anymore. If it does send those properties, they must be consistent.

### DIFF
--- a/src/backend/main/py.main/caliopen_main/discussion/store/discussion_index.py
+++ b/src/backend/main/py.main/caliopen_main/discussion/store/discussion_index.py
@@ -70,13 +70,16 @@ class DiscussionIndexManager(object):
         log.debug('No result found on index {}'.format(self.index))
         return {}, 0
 
-    def get_last_message(self, discussion_id, min_pi, max_pi):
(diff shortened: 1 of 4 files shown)


iamdey:
We need keep master stable so I think we should delay the merge of this branch since we will not fix the frontend bug you talk about until we have new specs from @LaurentChemla

Agree @sapiens-sapide  ?

sapiens-sapide:
I don't think we should postpone this PR because :  
 * This PR simplifies the way the frontend has to manage reply within discussion by handling 'default' replies, ie replies that do not change anything to the discussion (neither the subject, nor the participants.). Whatever specs from @LaurentChemla, Caliopen should handle a reply within a discussion including multiple participants, shouldn't it ?
* I investigated the frontend bug that prevent **some** replies to work properly and I concluded that the bug is not introduced by this PR but comes from an incorrect participants list handling from both backend and frontend. I need to investigate more on this bug, but it can be reproduced in current master branch.

PR #314, opened 2017-05-18
Refactor/frontend/pi

myleneb:
This adds a `mini` props to `<MultidimensionalPi>` and display `<MultidimensionalPi mini />` on messages.

By now: nothing on click or tap, more info is only displayed on hover, as a simple html `title="..."`+ css `cursor: help`. `Dropdown` could be used here for click/tap action, but maybe a kind of "tooltip" component would be better. @iamdey what's your opinion?

### DIFF
--- a/src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
+++ b/src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
@@ -41,7 +41,7 @@ const Ratings = ({ pi, piMax, averagePi, displayAveragePi, mini }) => {
   );
 
   return (
-    <div className={ratingsClassName}>
+    <div className={ratingsClassName} title={pi.map(p => ` ${p.name}: ${p.level}`)}>
       {displayAveragePi &&
         <Rating
           className="m-pi-ratings__item--average"
--- a/src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
+++ b/src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
@@ -2,26 +2,19 @@ import React, { PropTypes } from 'react';
 import classnames from 'classnames';
 import './style.scss';
 
-function classFor(element, mini) {
-  const classNames = [`m-pi-ratings__${element}`];
-  if (mini) { classNames.push(`m-pi-ratings--mini__${element}`); }
-
-  return classNames;
-}
-
 const Rating = ({ name, level, piMax, className, mini }) => {
   const width = (level / piMax) * 100;
   const style = { width: `${width}%` };
 
   return (
-    <div className={classnames(classFor('item', mini), className)}>
-      <div className={classnames(classFor('item-name', mini))}>
-        <span className={classnames(classFor('item-name-label', mini))}>{name}</span>
+    <div className={classnames('m-pi-ratings__item', { 'm-pi-ratings--mini__item': mini }, className)}>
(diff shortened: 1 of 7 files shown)


src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/style.scss
```diff
@@ -43,4 +45,27 @@ $gridColor: $co-color__fg__back--higher;
     padding-right: $co-component__spacing;
     text-transform: uppercase;
   }
+
+  &--mini {
+    max-width: $mini-pi__width;
+    cursor: help;
+
+    .m-pi-ratings__item {
```
iamdey: this should not be formated like this, it actually is what BEM tries to prevent.
Remember, when you see a DOM element with the class `m-pi-ratings--mini__item` and `m-pi-ratings__item` you will know what css directives you will find and what you can change without breaking other element design : there's no side-effects.

Long story short this should noted like this:

```
.m-pi-ratings {
  //...
  &--mini {
    // ...
    &__item {
      // ...
    }
    // ...
}
```

so in the template you have to declare it each time (and it is quite verbose but the safest way to know what happen):

`<item className={classnames('m-pi-ratings__item', 'm-pi-ratings--mini__item')} />`

That's the theory. In real world it is more pragmatic to leave it like this but it will never been fixed.

myleneb:
mini pi on message

src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
```diff
@@ -2,36 +2,41 @@ import React, { PropTypes } from 'react';
...
-  const ratingClassName = classnames(
-    'm-pi-ratings__item',
-    className,
-  );
 
   return (
-    <div className={ratingClassName}>
-      <div className="m-pi-ratings__item-name">
-        <span className="m-pi-ratings__item-name-label">{name}</span>
+    <div className={classnames(classFor('item', mini), className)}>
```
iamdey: well this is actually just fine of course :)

There is another way, and a bit more declarative : 

`className={classnames('m-pi-ratings__item`, { 'm-pi-ratings--mini__item': mini }, className)}`

it is not much more characters and it has the advantage to easily find a file containing the class (for example "m-pi-ratings--mini__item") when searching on the whole project in your IDE.

iamdey:
:+1:

src/frontend/web_application/src/components/MultidimensionalPi/components/Ratings/index.jsx
myleneb: I found the declarative way a bit confusing with all those looong classNames that's why I did a func for it. But I agree with your argument, this is not good practice for development.

LaurentChemla:
Peut-être prévoir un petit "?" clicable quelque part dans le dropdown pour ouvrir (dans une autre fenêtre ?) un texte explicatif des différents niveaux ?

myleneb:
oui @LaurentChemla !
Du coup:
- on peut créer la scene "Help" (ou "FAQ"), accessible par le menu User par ex., et qui contiendra toutes les infos d'aide,
- le "?" dans le dropdown du miniPi renverrait alors l'utilisateur vers la section "PI" de l'aide (dans nouvel onglet).
poke @Lauth0 @dey

iamdey:
Le sujet étant global, j'ai ajouté la story dans https://pad.caliopen.org/p/gestion-concep, il ne reste qu'à l’insérer dans la roadmap

PR #315, opened 2017-05-19
Feature/backend/handle large messages

sapiens-sapide:
This PR will add large emails handling capabilities to backend.  
Large messages body will be stored in an object storage service.  
For now, this service is a minio server, a S3 compatible object store.  
Minio has been added to the docker stack.

### DIFF
--- a/src/backend/brokers/go.emails/broker.go
+++ b/src/backend/brokers/go.emails/broker.go
@@ -68,6 +68,13 @@ func Initialize(conf LDAConfig) (broker *EmailBroker, connectors EmailBrokerConn
 			Hosts:       conf.StoreConfig.Hosts,
 			Keyspace:    conf.StoreConfig.Keyspace,
 			Consistency: gocql.Consistency(conf.StoreConfig.Consistency),
+			SizeLimit:   conf.StoreConfig.SizeLimit,
+		}
+		if conf.S3Service == "minio" {
+			c.WithS3 = true
+			c.Endpoint = conf.S3Config.Endpoint
+			c.AccessKey = conf.S3Config.AccessKey
+			c.SecretKey = conf.S3Config.SecretKey
 		}
 		b, e := store.InitializeCassandraBackend(c)
 		if e != nil {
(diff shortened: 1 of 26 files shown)


src/backend/brokers/go.emails/broker.go
```diff
@@ -68,6 +68,15 @@ func Initialize(conf LDAConfig) (broker *EmailBroker, connectors EmailBrokerConn
 			Hosts:       conf.StoreConfig.Hosts,
 			Keyspace:    conf.StoreConfig.Keyspace,
 			Consistency: gocql.Consistency(conf.StoreConfig.Consistency),
+			SizeLimit:   conf.StoreConfig.SizeLimit,
+		}
+		if conf.S3Service == "minio" {
```
gdchamal: this configuration apply to any s3 object store, minio or another.
Don't set explicit reference to minio

src/backend/configs/caliopen.yaml.template
```diff
@@ -18,6 +18,14 @@ broker:
     port: 2525
     host: localhost
 
+object_store:
+    db_size_limit: 1048576
+    service: minio
```
gdchamal: no explicit minio reference, it's an object store (s3 if you like) we don't care about implementation.

src/backend/main/py.main/caliopen_main/message/core/raw.py
```diff
@@ -35,17 +40,60 @@ def create(cls, raw):
...
+    def get(cls, raw_msg_id):
+        """
+        Get raw message from db or ObjectStorage service
+
+        :param raw_msg_id:
+        :return: a RawMessage or NotFound exception
+        """
+        try:
+            raw_msg = super(RawMessage, cls).get(raw_msg_id)
+        except NotFound:
```
gdchamal: except to raise the same ... let's remove this exception handler

src/backend/main/py.main/caliopen_main/message/core/raw.py
```diff
@@ -35,17 +40,60 @@ def create(cls, raw):
...
+                                    region=minioConf["raw_msg_location"])
+                try:
+                    resp = minioClient.get_object(url.netloc, path)
+                except ResponseError as exc:
+                    log.warn(exc)
+                    raise NotFound
+                # resp is a urllib3.response.HTTPResponse class
+                try:
+                    raw_msg.raw_data = resp.data
+                except Exception as exc:
```
gdchamal: Any exception to a NotFound. Previously an explicit ResponseError and no others Exception catched.

It should be more consistent, and may we may raise something else than Notfound hiding the real problem (that will eventually be a warn in a log file somewhere)

PR #320, opened 2017-05-23
Fix/frontend/import contacts

myleneb:
This improves ImportContactForm:
- creates ImportContact component (which contains *not-working* import function) in ContactBook,
- field error (file extension) but *not* global errors from response,
- fixes classes and variables names, improves CSS
- adds locales


src/frontend/web_application/src/components/ImportContactForm/presenter.jsx
```diff
@@ -1,139 +1,188 @@
...
 
 import './style.scss';
 
-const VALID_EXT = ['vcf', 'vcard'];
+const VALID_EXT = ['vcf', 'vcard']; // Valid file extensions for input#file
+
+const File = ({ file, onClick }) => (
+  <div className="m-import-contact-form__file">
+    <span className="m-import-contact-form__file-name">{file.name}</span>
+    <span className="m-import-contact-form__file-size">{file.size / 1000} ko</span>
```
iamdey: Size must be localized: ( FR: 1 Ko, EN: 1 KB )

`{ __('import-contact.file.size', { size: file.size / 1000 }) }`

and in locales:

`"import-contact.file.size": "%(size)s KB"`

`"import-contact.file.size": "%(size)s Ko"`

src/frontend/web_application/src/components/ImportContactForm/presenter.jsx
```diff
@@ -1,139 +1,188 @@
...
 import { FieldErrors } from '../form';
 import Button from '../Button';
 import Icon from '../Icon';
 
 import './style.scss';
 
-const VALID_EXT = ['vcf', 'vcard'];
+const VALID_EXT = ['vcf', 'vcard']; // Valid file extensions for input#file
+
+const File = ({ file, onClick }) => (
```
iamdey: I prefer something like `onRemove` instead of `onClick`

src/frontend/web_application/src/components/ImportContactForm/presenter.jsx
```diff
@@ -1,139 +1,188 @@
...
+
+  handleSubmitForm(ev) {
+    ev.preventDefault();
+    const { formData } = this.state;
+    this.props.onSubmit({ formData });
   }
 
-  onResetForm() {
+  resetForm() {
     document.getElementById('import-contact-form').reset();
```
iamdey: It works but it's not the "react way", I mean we should not use `document` or DOM accessors to do things when we have the choice.

If it's not possible to deal with this.state only, you can take a look to refs: https://facebook.github.io/react/docs/refs-and-the-dom.html

So you will be able to do something like `this.contactform.reset()`

src/frontend/web_application/src/components/ImportContactForm/presenter.jsx
```diff
@@ -1,139 +1,188 @@
...
     });
   }
 
-  validate(file) {
+  validateField(file) {
     const { __ } = this.props;
-    const error = __('Only .vcf or .vcard file.');
+    const error = __('import-contact.form.error.no_valid_ext');
     const ext = file.name ? file.name.split('.').pop() : null;
+    const formData = new FormData();
```
iamdey: same remark as above: `FormData` is part of DOM API, it doesn't work in node environment for instance.

In this case it should not be a pbm, but it might throw an error on SSR depending on the refactoring

myleneb: ok got it!
so a better way to deal with this form data it is to use an <iframe> with <form target="iframe" .. /> is that right? You will have to help for that because File upload seems very hard to me :D

src/frontend/web_application/src/components/ImportContactForm/presenter.jsx
myleneb: ok! this line was useless: no need to reset form because when state.file is null, it renders a form without any value in `<input>`.

src/frontend/web_application/src/components/ImportContactForm/presenter.jsx
iamdey: Yeah, file upload is hard in any way.

myleneb: btw what about [React-FileUpload](https://github.com/SoAanyip/React-FileUpload) or something similar? no way :D ?

PR #337, opened 2017-06-08
Upgrade stack to use Elasticsearch v5.4.1 (ie. current version).

sapiens-sapide:


sapiens-sapide:
Stack has been tested locally and within docker with ES 5.4.1.
Everything seems OK : setup db, create user, import and basic usage.

PR #344, opened 2017-06-15
Fix/frontend/contact profile

myleneb:
This fixes ContactProfile and ContactDetails according to UX.
- adds components OrgaForm and OrgaDetails, IdentityForm and IdentityDetails
- adds tags and `<MultidimensionalPi />`  on `<ContactDetails />`
- adds fields on `<ContactDetailsForm />`
- adds editMode for each panel on `<ContactDetails />`
- fix responsive avatar size
- adds locales

`<Contact />` is mostly a clone from `<Account />` without anything related to user.

Edit one ContactDetails panel on desktop:

Edit profile on mobile:


src/frontend/web_application/src/components/ContactDetails/components/OrgaDetails/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  render() {
+    const { organization, editMode } = this.props;
+
+    return (
+      <ItemContent large key={organization.organization_id}>
+        <Icon type="briefcase" className="m-orga-details__icon" />
+        {organization.job_description}
+        {organization.job_description && organization.name && ' at '}{organization.name}
```
iamdey:
```jsx
renderJobTitle(organization) {
  const departement = organization.department ? ` (${departement})` : '';
  if (organization.job_description && organization.name) {
    return __('orga-details.job.desc-full', { jobDesc: organization.job_description, orgaName: organization.name, departement });
  }

  return (organization.job_description || organization.name) && departement;
}
```
```json
{
 "orga-details.job.desc-full": "%(jobDesc)s at %(orgaName)s%(departement)s"
}
```

?

src/frontend/web_application/src/components/ContactDetails/components/OrgaForm/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+import { FieldErrors, Fieldset, Legend, TextFieldGroup, FormGrid, FormRow, FormColumn } from '../../../form';
+
+import './style.scss';
+
+
+class OrgaForm extends Component {
+  static propTypes = {
+    errors: PropTypes.arrayOf(PropTypes.string),
+    onSubmit: PropTypes.func.isRequired,
+    __: PropTypes.func.isRequired,
```
iamdey: in case of edition, the original orga is missing, isn't it ? 

You might need to use the `generateStateFromProps` pattern ?

src/frontend/web_application/src/components/ContactDetails/components/OrgaForm/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+  static defaultProps = {
+    errors: [],
+  };
+
+  constructor(props) {
+    super(props);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleInputChange = this.handleInputChange.bind(this);
+    this.state = {
+      orgaDetails: {
```
iamdey: may be stick to the same name as the props: `organization`

src/frontend/web_application/src/components/ContactDetails/components/OrgaForm/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+          <Legend>
+            <Icon className="m-orga-form__icon" type="building" />
+            {__('contact.orga_form.legend')}
+          </Legend>
+          <FormRow>
+            {errors.length > 0 && (<FormColumn><FieldErrors errors={errors} /></FormColumn>)}
+            <FormColumn>
+              <TextFieldGroup
+                name="name"
+                type="text"
```
iamdey: type is required ?

src/frontend/web_application/src/components/ContactDetails/index.jsx
```diff
@@ -108,7 +124,7 @@ class ContactDetails extends Component {
           email={email}
           allowConnectRemoteEntity={allowConnectRemoteEntity}
           remoteIdentity={remoteIdentity}
-          editMode={this.state.editMode}
```
iamdey: I like to stick to full notation: `this.state.editMode` because it helps visually to distinguish what is props and what is state.

eg.

```
render() {
  const { foo } = this.props;

  return (<div><p>{foo}</p><span>({this.state.bar})</span></div>);
}
```

src/frontend/web_application/src/components/ContactDetails/index.jsx
```diff
@@ -182,21 +203,62 @@ class ContactDetails extends Component {
...
+                      editMode={editMode.organizations}
+                      onDelete={onDeleteContactDetail}
+                      __={__} />
+                  )}
+                  </TextList>
+                {editMode.organizations && <OrgaForm onSubmit={onAddContactDetail} __={__}/>}
+              </div>
+          </div>
+        }
+        {contact.identities !== [] &&
```
iamdey: I'm not sure if identities is always defined in contacts
(API might not send props which are empty)

```
{contact.identities && contact.identities.length > 0 && ( ...
```

src/frontend/web_application/src/components/Icon/index.jsx
```diff
@@ -76,7 +78,7 @@ Icon.propTypes = {
 };
 Icon.defaultProps = {
   className: undefined,
-  type: undefined,
+  type: false,
```
iamdey: we defined `type` as a string : when giving default bool it is not coherent in my mind (and I guess react throws a warning).

I think undefined is just fine : we assume it is not set and we can deal with it with a condition same like `false`

src/frontend/web_application/src/scenes/Contact/presenter.jsx
```diff
@@ -0,0 +1,136 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      editMode: false,
+      isFetching: true,
+    };
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
```
iamdey:
```
const { requestContact, contactId } = this.props;
requestContact({ contactId });
```

but I can take care of the redux part if you want.

src/frontend/web_application/src/components/Icon/index.jsx
myleneb: Ok sorry this comes from a very bad file manipulation from me. I didn't edit this, I copied the file from an old branch before: https://github.com/CaliOpen/Caliopen/commit/cf1c15f3423a4317e2a48b28ad40dda8d00d0f5e 
when I didn't know how to revert a commit (... I know it's not good)

btw of course I understand `undefined` match perfectly this prop :)

src/frontend/web_application/src/components/ContactDetails/index.jsx
```diff
@@ -74,22 +87,22 @@ class ContactDetails extends Component {
...
   }
 
-  renderSubtitleActions() {
+  renderSubtitleActions(panel) {
     const { __ } = this.props;
-    const activeButtonProp = this.state.editMode ? { color: 'active' } : {};
+    const activeButtonProp = this.state.editMode[panel] ? { color: 'active' } : {};
 
     return (
       <Button
```
myleneb: @ dey there is a bug here: Button doesn't responds to click when you click on the inner `<Icon />`. If you click anywhere else on the Button, it works. I played with "pointer-event" css on the Icon, but doesn't solve the issue.
I think it's because `value` attribute on Button is not null (that's the only difference with other Buttons that work fine), but can't fix that.

src/frontend/web_application/src/styles/object/o-avatar.scss
```diff
@@ -2,7 +2,8 @@ $m-avatar__size: (
   small:  2rem,
   medium: 3rem,
   large:  4rem,
-  xlarge: 6rem
+  xlarge: 6rem,
+  xxlarge: 8rem
```
myleneb: From a technical point of view I agree, 3 sizes should be enough.
from a graphical point of view, 1rem can make the difference ^^

iamdey: I agree with you. 

May be the problem is the notation: `small, xxlarge ... ` are not good at the end if we have more than 3 sizes. 

This can be a good case for http://cssinjs.org on the other side extending styles can be a bad thing since allows inconsistent designs.

src/frontend/web_application/src/components/ContactDetails/index.jsx
iamdey: ok I understand ... so it's sure, we can't use "value" property in the button, it's not its purpose (since there is no props `value` in Button component).

I recently got what you wanted to do with [currying](https://medium.com/@kbrainwave/currying-in-javascript-ce6da2d324fe) so "value" is no more required:

```diff
...
- handleSwitchEditMode() {
+ makeHandleSwitchEditMode(panel) {
+   return () => {
      this.setState(prevState => ({
-       editMode: !prevState.editMode,
+       editMode: {
+         [panel]: !prevState.editMode[panel],
+       }
     }));
+  };
}
...
- onClick={this.handleSwitchEditMode}
+ onClick={this.makeHandleSwitchEditMode(panel)}
```

(actually it's not currying as described in the article but it's the same result)

I hope this fix your issue

src/frontend/web_application/src/components/ContactDetails/components/OrgaForm/index.jsx
myleneb: @iamdey don't you think the purpose of OrgaForm is to add a new organization, so the original organization doesn't have to appear here? Original organizations are returned in OrgaDetails, with remove Button on editMode (just like other forms on the page)

iamdey: As a user I think, I'll be frustrated if I have to delete then re-add a whole organization just because of a typo to fix.

PR #347, opened 2017-06-15
Privacy policy on Signup form

myleneb:
This adds a short privacy policy text + checkbox on Signup form.
"More info" link opens Modal witch contains .. more info about Piwik use on Caliopen :)


src/frontend/web_application/src/components/SignupForm/presenter.jsx
```diff
@@ -59,6 +66,19 @@ class SignupForm extends Component {
     this.setState(generateStateFromProps(newProps));
   }
 
+  handleOpenModal() {
+    this.setState(prevState => ({
```
iamdey: if you don't need prevState for specific condition, you can omit it:

```
this.setState({ isModalOpen: true });
```

This will not affect the rest of the state (yes I know it's a bit weird).

src/frontend/web_application/src/components/SignupForm/presenter.jsx
```diff
@@ -186,6 +232,31 @@ class SignupForm extends Component {
             </FormColumn>
           </FormRow>
           <FormRow>
+            <FormColumn bottomSpace>
+              <div className="s-signup__privacy">
+                <h4>{__('signup.form.privacy.title')}</h4>
+                <p className="s-signup__privacy__text">
+                  {__('signup.form.privacy.intro')}
+                  <PrivacyLink
```
iamdey: no need to create a new component I think (and I don't like much create a new component at runtime, I mean directly in `render`)

PR #349, opened 2017-06-16
Feature/backend/download attachments

sapiens-sapide:
Implements route `GET /v2/messages/{message_id}/attachments/{attachment_index}`


src/backend/defs/go-objects/attachment.go
```diff
@@ -9,5 +9,5 @@ type Attachment struct {
 	File_name    string `cql:"file_name"        json:"file_name"`
 	Is_inline    bool   `cql:"is_inline"        json:"is_inline"`
 	Size         int    `cql:"size"             json:"size"`
-	URI          string `cql:"uri"              json:"uri"` // ObjectStore url for temporary file (draft)
+	URI          string `cql:"uri"              json:"uri"` // ObjectStore url for temporary file (draft) or boundary reference for mime-part attachment
```
gdchamal: Why not use a specific field rather than reuse URI ?

sapiens-sapide: Because URI is exactly for this kind of situation : locate a ressource whatever the location and the protocol is. If the attachment is a temporary file, URI will be in the form of `s3://path/to/file` ; if attachment is embedded in the raw message, URI will be in the form of `urn:email:mime-boundary:--part-boundary--`.  
(Note that for now, URI is only use to double check that we found the right attachment, because we already know where to seek the attachment by only checking if the message is a draft or not.)

Issue #345, opened 2017-06-15
unable to upload a vcard file


iamdey:
when trying to upload using a relly basic html form with an input I got the ERR422 with this log:

```
apiv1_1            | 2017-06-15 08:20:50,027 ERROR [caliopen_api.user.imports][waitress] File valid but we can create the new contact: At line 4: Failed to parse line: -----------------------------108426991410396656342064448763
```

I use the following html: 

```
    <form action="http://localhost:31415/api/v1/imports" method="post" enctype="multipart/form-data">
      <label>File <input type="file" name="file" /></label>
      <button type="submit">Submit</button>
    </form>
```

_(the request must be tricked to pass the Authorization Bearer)_

Here is the request headers:

```
Host: localhost:31415
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:45.0) Gecko/20100101 Firefox/45.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
DNT: 1
Authorization: Bearer Yzk2NTNkYjMtZTYxOC00ZGRmLThlN2EtNzYwODU4YmVhNzY4OmEwNTkyYTQ4NTMxNzEyMGMyNjRkNzI1ZTFhM2I1NDAxNzY5MjU2NDU=
Referer: http://localhost:31415/api/v1/imports
Content-Length: 444
Content-Type: text/plain;charset=UTF-8
```

body:

```
Content-Type: multipart/form-data; boundary=---------------------------13924900811460510303354933696
Content-Length: 319

-----------------------------13924900811460510303354933696
Content-Disposition: form-data; name="file"; filename="fake.vcf"
Content-Type: text/vcard

BEGIN:VCARD
VERSION:3.0
FN:Fake
N:;Fake;;;
NICKNAME:f4ke
TEL;TYPE=CELL:+33654321000
END:VCARD

-----------------------------13924900811460510303354933696--
```

---

Well now, I prefer to use axios without basic form since it uses iframe and old stuff like that.
(but we still might need this old fashion way to support more browser -> after alpha)

The request headers & body will be a bit different, `Content-Type` and `boundary` are in request headers (instead of body headers) and so body has only data:

```
-----------------------------13757638558450105911532952306
Content-Disposition: form-data; name="file"; filename="fake.vcf"
Content-Type: text/vcard

BEGIN:VCARD
[...]

PR #351, opened 2017-06-16
Fix/backend/345

lisonferez:
- Resolve conflicts in imports.py (import logging and log = logging.getLogger(__name__) twice on the file).
- Fix problem with validate phone number (01 02 03 04 05 incorrect phone number for example)

### DIFF
--- a/src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
+++ b/src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
@@ -9,9 +9,6 @@ from pyramid.response import Response
 from caliopen_main.parsers.vcard import parse_vcards
 from caliopen_main.parsers.vcard import read_file
 
-import logging
-log = logging.getLogger(__name__)
-
 from ..base.exception import (ValidationError,
                               Unprocessable)
 
--- a/src/backend/main/py.main/caliopen_main/parsers/vcard.py
+++ b/src/backend/main/py.main/caliopen_main/parsers/vcard.py
@@ -48,7 +48,8 @@ def parse_vcard(vcard):
         if vcard.contents['n'][0].value.additional:
             for a in vcard.contents['n'][0].value.additional:
                 if len(a) == 1:
-                    new_contact.additional_name = vcard.contents['n'][0].value.additional
+                    tmp = vcard.contents['n'][0].value.additional
+                    new_contact.additional_name = tmp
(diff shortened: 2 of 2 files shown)


src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
```diff
@@ -9,9 +9,6 @@
 from caliopen_main.parsers.vcard import parse_vcards
 from caliopen_main.parsers.vcard import read_file
 
-import logging
-log = logging.getLogger(__name__)
```
iamdey: sure ? Does `log` at line 17 is global ?

src/backend/main/py.main/caliopen_main/parsers/vcard.py
```diff
@@ -193,6 +193,18 @@ def parse_vcard(vcard):
...
+        elif v == 'tel':
+            for tel in vcard.contents['tel']:
+                phone = NewPhone()
+                phone.is_primary = False
+                try:
+                    func = PhoneNumberType.validate_phone
+                    phone.number = func(PhoneNumberType(),tel.value)
+                    new_contact.phones.append(phone)
+                except:
+                    pass
```
iamdey: does it mean that a phone is not well formed it will be drop ?

for e.g: `01 02 03 04 05` is not "valid" cf. #345 , so it will not be saved ?
Is there any other way ?

may be a normalizer or whatever ? Or at least save in custom fields ?

iamdey: It's ok, If we don't know yet how to deal with that but we can run a process afterward only if with the information : for example a script that retrieve the country of connection et so we will be able to propose a region fix for "bad" numbers.

gdchamal: For the moment invalid phone number (not international format) will be dropped.

We need to define a strategy to guess a country related to a phone number for the normalization library we used able to process.

We can relax condition and store phone number even if it can't be normalized but that's not goal for long term

src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
gdchamal: import is done line 1 and log instanciated line 20 on master. Sound legit to remove this doublon

PR #352, opened 2017-06-16
Remove all references to user Counter model

gdchamal:
Cleanup completly no more used model user counters, as they are deprecated in cassandra and not supported at all by scylladb.


sapiens-sapide:
I saw a `rule_id` within `caliopen_main/user/store/user.py` that seems to be a reference to counter.rule_id.  

Do we need to remove this property ?

gdchamal:
Nice catch. Will move this user.rule_id to an UUID type as all others resources are identified

PR #355, opened 2017-06-19
[wip]Feature/backend/prepare remote identities

gdchamal:
Rework RemoteIdentity to an usable form for fetching an IMAP source.

Make a basic API to create a remote identity for an user.

### DIFF
--- a/src/backend/main/py.main/caliopen_main/user/core/user.py
+++ b/src/backend/main/py.main/caliopen_main/user/core/user.py
@@ -383,3 +383,21 @@ class User(BaseCore):
         except Exception as exc:
             log.error('Unexpected exception {}'.format(exc))
         return False
+
+    def add_remote_identity(self, remote):
+        """Add a remote identity to an user.
+
+        return: the RemoteIdentity created instance.
+        rtype: RemoteIdentity
+        """
+        # XXX check for duplicate before insert
+        identifier = remote.identifier.lower()
+        name = remote.display_name if remote.display_name else identifier
+        identity = RemoteIdentity.create(user_id=self.user_id,
+                                         identifier=identifier,
+                                         display_name=name,
+                                         type=remote.type,
+                                         status=remote.status,
+                                         last_check=None,
+                                         infos=remote.infos)
+        return identity
--- a/src/backend/main/py.main/caliopen_main/user/core/user.py
+++ b/src/backend/main/py.main/caliopen_main/user/core/user.py
@@ -20,8 +20,8 @@ from ..store import (User as ModelUser,
                      FilterRule as ModelFilterRule,
                      ReservedName as ModelReservedName,
                      LocalIdentity as ModelLocalIdentity,
-                     ContactLookup
-                     )
+                     RemoteIdentity as ModelRemoteIdentity,
+                     ContactLookup)
 
 from caliopen_storage.core import BaseCore, BaseUserCore, core_registry
 from .contact import Contact as CoreContact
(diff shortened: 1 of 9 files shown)


src/backend/main/py.main/caliopen_main/user/core/user.py
```diff
@@ -383,3 +383,21 @@ def add_local_identity(self, address):
...
             log.error('Unexpected exception {}'.format(exc))
         return False
+
+    def add_remote_identity(self, remote):
+        """Add a remote identity to an user.
+
+        return: the RemoteIdentity created instance.
+        rtype: RemoteIdentity
+        """
+        # XXX check for duplicate before insert
```
sapiens-sapide: Is it a todo ? (because I don't see the duplicate checking in current code)

src/backend/main/py.main/caliopen_main/user/store/user.py
```diff
@@ -74,11 +74,12 @@ class RemoteIdentity(BaseModel):
...
 
     user_id = columns.UUID(primary_key=True)
-    identity_id = columns.Text(primary_key=True)
+    identifier = columns.Text(primary_key=True)
+    display_name = columns.Text()
     type = columns.Text()
     status = columns.Text()
-    credentials = columns.List(columns.Text)
     last_check = columns.DateTime()
+    infos = columns.Map(columns.Text, columns.Text)
```
sapiens-sapide: Is this field for storing credentials ? Or do we need to add 2 more fields : `username`, `password`.  
BTW, we need to choose a more secure storage strategy for remote identities credentials.

src/backend/main/py.main/caliopen_main/user/core/user.py
gdchamal: yes I use XXX to identify part of code to enhance later

src/backend/main/py.main/caliopen_main/user/store/user.py
gdchamal: username/password is not an universal scheme (could be a client_secret and client_token on an oauth authentication system), so I prefer to be agnostic about informations for collecting a remote identity. informations stored are dependent of remote identity type (server, port, username, password for an imap).

About storing this credentials encrypted, I agree completely. But we must define how to do (which encryption/decryption algorithm, how to decrypt, etc). For development purposes this is not needed now and we can enhance later with encryption when we have define how to proceed

sapiens-sapide: OK. Could we have a more explicit field name, instead of `infos` ? : 
* credentials ?
* account_settings ?
*…

gdchamal: settings ?

sapiens-sapide: I'd preferred a field name that suggest sensitive informations.

gdchamal: OK in fact I'll integrate vault service (by hashicorp) to store these sensitives informations.

In RemoteIdentity this field is no more needed, the related informations/credentials,server,port,etc will be stored in vault under a remote_id key

WDYT ? we've got a way to secure correctly such sensitive informations

sapiens-sapide: great !

gdchamal:
Integrate hashicorp vault in the stack for storing of remote identities connection informations.

I was thinking such vault integration later, but it could be a cool benefit not so hard to gain

gdchamal:
Some difficulties using easily vault in docker stack for the moment. Merge in this state will come back later for such integration

PR #357, opened 2017-06-21
Upgrade/backend/elasticsearch

gdchamal:
Using official elasticsearch v5.x image need usage of docker volumes (https://docs.docker.com/engine/reference/commandline/volume_create/#parent-command)

A previous merge to upgrade to es 5.4.1 forgot that needed update and make es container does not work anymore on linux platform with our volume usage in docker compose.

### DIFF
--- a/devtools/docker-compose.yml
+++ b/devtools/docker-compose.yml
@@ -68,6 +68,7 @@ services:
       environment:
           - xpack.security.enabled=false
           - xpack.monitoring.enabled=false
+          - "ES_JAVA_OPTS=-Xms512M -Xmx512M"
       volumes:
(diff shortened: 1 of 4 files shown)


gdchamal:
Need to tune a bit linux host kernel issuing this command:

```
sudo sysctl -w vm.max_map_count=262144
```

in case elasticsearch container does not start with this type of error:
https://github.com/docker-library/elasticsearch/issues/111

sapiens-sapide:
LGTM. (re)-tested on OSX.

iamdey:
how do we reset datas with this config ?

sapiens-sapide:
resetting data is documented in another branch : `doc/backend/docker-stack`

PR #363, opened 2017-06-27
Feature/frontend/piwik react router

myleneb:
Basic install of piwik-react-router (https://github.com/joernroeder/piwik-react-router), and checkbox component on Signup form (+ info text).


gdchamal:
I'm not sure about the last sentence. At this time piwik data are anonymized (IP address truncated) but it's a configuration for the public site.

We will need individual tracking so it can't be anonymous data collected.

BTW "tracke" in french it's not OK, collecte ?

src/frontend/web_application/src/index.jsx
```diff
@@ -2,11 +2,19 @@ import React from 'react';
...
+import { v1 as uuidv1 } from 'uuid';
 import App from './App';
 import configureStore from './store/configure-store';
 import getRouterHistory from './services/router-history';
 import { getLocale } from './services/i18n';
 
+const piwik = PiwikReactRouter({
+  url: 'https://piwik.caliopen.org/analytics',
+  siteId: 6,
+  userId: uuidv1(), //userId can be username, email adress or uuid
```
iamdey: I don't know if this will work well. May be it's enough : the userId is generated only one time by "browser session".

But we will miss the association with real user. May be it can be done only if we can update the piwik's configuration on the fly (when user is authenticated, I mean).

iamdey:
:+1: go merge

Issue #364, opened 2017-06-27
Patch contact


iamdey:
ERR 400 on contact PATCH

* [ ] missing id
* [ ] additional params

---

missing id : Emails / addresses:

request payload:
```
{
	"current_state": {
		"emails": [{
			"email_id": "d56b19de-3e11-4c19-8a63-fa7224915a4f",
			"is_primary": false,
			"type": "other",
			"address": "alan@idiocy.org"
		}]
	},
	"emails": [{
		"email_id": "d56b19de-3e11-4c19-8a63-fa7224915a4f",
		"is_primary": false,
		"type": "other",
		"address": "alan@idiocy.org"
	}, {
		"address": "aazd@azdkh",
		"type": "work",
		"is_primary": false
	}]
}
```

response:

```
{
	"errors": [{
		"code": 400,
		"description": "The server could not comply with the request since it is either malformed or otherwise incorrect.",
		"component": "Unknown",
		"values": 400,
		"message": "u'email_id' is a required property\n\nFailed validating u'required' in schema[u'properties'][u'emails'][u'items']:\n    {u'additionalProperties': False,\n     u'properties': {u'address': {u'type': u'string'},\n                     u'email_id': {u'type': u'string'},\n                     u'is_primary': {u'type': u'boolean'},\n                     u'label': {u'type': u'string'},\n                     u'type': {u'type': u'string'}},\n     u'required': [u'email_id', u'address'],\n     u'type': u'object'}\n\nOn instance[u'emails'][1]:\n    {u'address': u'aazd@azdkh', u'is_primary': False, u'type': u'work'}",
		"property": "RequestValidationError",
		"type": "Bad Request"
	}]
}
```

iamdey:
Phones:

request payload:

```
{
	"current_state": {},
	"phones": [{
		"number": "654",
		"type": "work",
		"is_primary": false
	}]
}
```

response:

```
{
	"errors": [{
		"code": 400,
		"description": "The server could not comply with the request since it is either malformed or otherwise incorrect.",
		"component": "Unknown",
		"values": 400,
[...]

PR #366, opened 2017-06-28
Fix/backend/swagger definitions

sapiens-sapide:
Fix issue #364.  

We need another fix to correctly handle patches that add new nested objects. Will do it in a separate branch.

### DIFF
--- a/src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
+++ b/src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
@@ -19,13 +19,19 @@ log = logging.getLogger(__name__)
 
 @resource(collection_path='/imports', path='')
 class ContactImport(Api):
-
     def __init__(self, request):
         self.request = request
         self.user = request.authenticated_userid
 
     @view(permission='authenticated')
     def collection_post(self):
+
+        # need to check by ourself if <file> param is present
+        # because swagger lib failed to do it correctly :(
+        try:
+            self.request.POST.getone("file")
+        except:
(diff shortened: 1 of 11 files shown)


src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
```diff
@@ -19,14 +19,20 @@
...
 
     @view(permission='authenticated')
     def collection_post(self):
 
+        # need to check by ourself if <file> param is present
+        # because swagger lib failed to do it correctly :(
+        try:
+            self.request.POST.getone("file")
+        except:
+            raise ValidationError("param <file> is missing in form-data")
```
gdchamal: should be better to raise a correct HTTP error code (HTTPBadRequest)

src/backend/main/py.main/caliopen_main/objects/base.py
```diff
@@ -369,9 +369,12 @@ def apply_patch(self, patch, **options):
             msg = "Patch current_state not consistent with db, step {} key {}"
             if key not in patch_current.keys():
                 # means patch wants to add the key. Value in db should be null
+                # TODO : need to check if cur_val is an empty array or empty dict
+                print(cur_val)
```
gdchamal: no print please

src/backend/interfaces/REST/py.server/caliopen_api/user/imports.py
sapiens-sapide: ValidationError is a HTTPBadRequest

PR #368, opened 2017-06-29
Fix/backend/nested objects patch creation

sapiens-sapide:


src/backend/main/py.main/caliopen_main/objects/base.py
```diff
@@ -419,6 +365,62 @@ def apply_patch(self, patch, **options):
...
+        else:
+            if isinstance(current_attr, types.ListType):
+                if old_val == [] and cur_val != []:
+                    return main_errors.PatchConflict(
+                        message=msg.format(1, key))
+                if cur_val == [] and old_val != []:
+                    return main_errors.PatchConflict(
+                        message=msg.format(2, key))
+                for old in old_val:
+                    found = False
```
gdchamal: You can use the for: else syntax instead of using a found variable

sapiens-sapide: Like this : 16DB822

Issue #369, opened 2017-06-29
Err 422 : expected string or buffer


iamdey:
when trying to patch a draft message (new discussion):

```
{
	"current_state": {
		"participants": [{
			"type": "To",
			"label": "dey+1@bitbucket.local",
			"protocol": "unknown",
			"address": "dey+1@bitbucket.local"
		}, {
			"protocol": "email",
			"type": "From",
			"label": "None None",
			"contact_ids": ["bf5b8211-e85b-4c18-a346-427482ea98f4"],
			"address": "foo@caliopen.local"
		}]
	},
	"participants": [{
		"address": "test1@bitbucket.local",
		"protocol": "email",
		"label": "test1@bitbucket.local",
		"type": "To"
	}]
}
```

response:

```
{
	"errors": [{
		"code": 422,
		"description": "Application encountered an error when applying patch",
		"component": "message",
		"values": 422,
		"message": "expected string or buffer",
		"property": "MergePatchError",
		"type": "Patch Error"
	}]
}```

**Comment by iamdey** [comment]
I still have this issue on an old message. I'll try on a new one

**Comment by iamdey** [comment]
I have a new conversion & message with no participants, I try to add one:

req:

```
{
	"current_state": {
		"participants": [{
			"protocol": "email",
			"type": "From",
			"label": "None None",
			"contact_ids": ["bf5b8211-e85b-4c18-a346-427482ea98f4"],
			"address": "foo@caliopen.local"
		}]
	},
	"participants": [{
		"address": "dey@inbucket.local",
		"protocol": "email",
		"label": "dey@inbucket.local",
		"type": "To"
	}]
}
```

resp:

```
{
	"errors": [{
		"code": 422,
		"description": "Application encountered an error when applying patch",
		"component": "message",
		"values": 422,
		"message": "expected string or buffer",
		"property": "MergePatchError",
		"type": "Patch Error"
	}]
}
```

iamdey:
PS: when posting new message with participants, it's ok but it fails  when trying to edit message body

iamdey:
It's ok now

PR #371, opened 2017-06-29
Fix/backend/patch draft

sapiens-sapide:
Fix for issue #369 + patch improvements.


src/backend/main/py.main/caliopen_main/message/parameters/draft.py
```diff
@@ -90,13 +86,19 @@ def _add_from_participant(self, user_id):
         user = User.get(user_id)
         if not hasattr(self, 'participants'):
             self.participants = []
+        else:
+            if len(self.participants) > 0:
+                for i, participant in enumerate(self.participants):
+                    if re.match("[Ff][Rr][Oo][Mm]", participant.type):
```
gdchamal: a re.match("from", participant.type, re.I) should be more lisible (re.I for ignore case option)

PR #375, opened 2017-06-30
Add MenuBar to Contact

myleneb:
This adds MenuBar to Contact (and fix negative margin on top of page)
- MenuBar with action Buttons (Delete user, Edit tags) to fit UX
- Modal for tags
- CSS for Account, ContactProfile and ContactDetails
- locales


src/frontend/web_application/src/scenes/Contact/presenter.jsx
```diff
@@ -34,19 +50,65 @@ class Contact extends Component {
...
+  }
+
+  handleCloseTagsModal() {
+    this.setState({ isTagModalOpen: false });
+  }
+
+  renderTagsModal() {
+    const { contact, updateContact, __ } = this.props;
+    const count = contact.tags ? contact.tags.length : 0;
+    const title = [
```
iamdey: I think this works perfectly but I do not like the `key="1"` to prevent warning/error about array of nodes.

what about something more straight like:

```jsx
(<span>__('tags.header.title') <span className="m-tags-form__count">{__('tags.header.count', { count }) }</span></span>)
```

src/frontend/web_application/src/scenes/Contact/presenter.jsx
```diff
@@ -34,19 +50,65 @@ class Contact extends Component {
...
+              responsive="icon-only"
+              icon="trash"
+            >{__('contact.action.delete_contact')}</Button>
+            <Button
+              onClick={this.handleClickTags}
+              responsive="icon-only"
+              icon="tags"
+            >{__('contact.action.edit_tags')}</Button>
+            {this.renderTagsModal()}
+            </MenuBar>
```
iamdey: not big deal but ident is incorrect at this line and the one after

myleneb: ok seems my linter is not working properly since latest atom update :/

PR #384, opened 2017-07-03
fix elasticsearch error permanently

iamdey:


devtools/README.md
```diff
@@ -141,4 +141,7 @@ According to [this issue](https://github.com/docker-library/elasticsearch/issues
 
 ```
 sysctl -w vm.max_map_count=262144
+
+# make it permanent:
+echo "vm.max_map_count = 262144" > sudo tee /etc/sysctl.d/vm.caliopen.conf
```
iamdey: does this work with major linux distributions ?

gdchamal: yes AFAIK

Issue #386, opened 2017-07-03
docker containers expose port to host machine


gdchamal:
We must NOT expose container ports to host interface, they will be visible from outside. It leads to problems and bad solutions (iptables management for ansible deployment).

For poc usage it sound that we must use expose and not ports compose keyword for such restriction: https://stackoverflow.com/questions/35429837/docker-compose-port-mapping#35441128

PR #387, opened 2017-07-04
add docker-compose configuration for production

iamdey:
* ports will not be published on host
* frontend (4000) and broker (2525) are publihed to 127.0.0.1
* add doc for production scenarios

fix: #386

### DIFF
--- a/devtools/README.md
+++ b/devtools/README.md
@@ -142,3 +142,16 @@ According to [this issue](https://github.com/docker-library/elasticsearch/issues
 ```
 sysctl -w vm.max_map_count=262144
 ```
+
+# Staging scenarios
+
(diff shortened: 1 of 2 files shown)


iamdey:
When stack is started, I get this:

```
netstat -pln | grep docker
tcp        0      0 127.0.0.1:2525          0.0.0.0:*               LISTEN      2698/docker-proxy
tcp        0      0 127.0.0.1:4000          0.0.0.0:*               LISTEN      1522/docker-proxy
```

gdchamal:
I still not confident at all with this "production" naming.

It's not a real production platform in term of infrastructure. stage fit better my definition of what is such platform

Issue #341, opened 2017-06-14
contact book - duplicated alphabetic group 


iamdey:
Don't know if it is due to back or front or inconsistent data (cf. 2nd screenshot)

![capture d ecran de 2017-06-14 11-33-08](https://user-images.githubusercontent.com/65737/27125727-852bd27a-50f5-11e7-84f1-a6b9cfd497e4.png)

![capture d ecran de 2017-06-14 11-32-58](https://user-images.githubusercontent.com/65737/27125726-839b79ce-50f5-11e7-80e7-913eb422a0b5.png)

myleneb:
yes: as `contact.title` is null for this entry, the function that sorts contacts by first letters bugs (I didn't found how to fix it on frontend).
This shouldn't happen because `contact.title` will never, never be null, according to @gdchamal

iamdey:
but happened :D

gdchamal:
In fact there is only one contat with title null due to malformed fixture.
I'll fix import function to not create contact if title is null

iamdey:
I think the actual contact with bad fixture is the user

iamdey:
title is now always given. bug doesn't appear anymore

PR #389, opened 2017-07-04
Fix/frontend/user in contactbook

myleneb:
This fix issue https://github.com/CaliOpen/Caliopen/issues/341 and also fix a bug in TagList (TagList should not display blank tags)


iamdey:
My thought is we should not fix those errors, it is backend responsibility:

1. I think user's contact profile should be in contactbook (I can send a message to my future myself); 
    routes /settings/me-contact & /contacts/<my contact id> give almost same page
    If `title` is missing it's a bug of the backend
2. About tag it's almost same thing, an empty tag should not exists, it's a bug of the backend
    May be it's not a bug so we might need clarify the usage.

iamdey:
1. the bug #341 should be fixed in backend
2. tag should be empty #416

Imo you can close this PR

PR #405, opened 2017-07-20
Add country and region selectors

myleneb:
- install react-country-region-selector
- update styles
- update locales

### DIFF
--- a/src/frontend/web_application/src/components/ContactDetails/components/AddressForm/index.jsx
+++ b/src/frontend/web_application/src/components/ContactDetails/components/AddressForm/index.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { CountryDropdown, RegionDropdown } from 'react-country-region-selector';
 import Icon from '../../../Icon';
 import Button from '../../../Button';
 import { FieldErrors, Fieldset, Legend, TextFieldGroup, SelectFieldGroup, FormGrid, FormRow, FormColumn } from '../../../form';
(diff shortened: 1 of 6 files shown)


src/frontend/web_application/src/components/ContactDetails/components/AddressForm/index.jsx
```diff
@@ -20,8 +21,6 @@ class AddressForm extends Component {
 
   constructor(props) {
     super(props);
-    this.handleSubmit = this.handleSubmit.bind(this);
-    this.handleInputChange = this.handleInputChange.bind(this);
     this.state = {
       contactDetail: {
```
ziir: You might want to keep the `state` as flat as possible, remove all unnecessary namespaces

This component is good example because the only thing stored in the Component's state are "contact details", thus the `contactDetails` namespace is unnecessary and gets in the way of potential optimizations

src/frontend/web_application/src/components/ContactDetails/components/AddressForm/index.jsx
```diff
@@ -20,8 +21,6 @@ class AddressForm extends Component {
...
         street: '',
```
ziir: Not initializing the state from props?

src/frontend/web_application/src/components/ContactDetails/components/AddressForm/index.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { CountryDropdown, RegionDropdown } from 'react-country-region-selector';
```
ziir: `react-country-region-selector` is ~83kB minified, it's good candiate for both async-loading and some webpack-magic

PR #406, opened 2017-07-21
Fix `Unknown props` error on Dropdown

myleneb:
as you can't pass custom props to `<div>`, you have to list them before `...props`


myleneb:
@iamdey @ziir I don't understand why this PR doesn't pass tests... can you help?

iamdey:
:+1: for the PR

About the test: I think this is still unstable.
You can try to login to Travis then restart the build.
I'm OK if you merge it in this state but if you want to make sure your dev
doesn't brake anything you can run functional tests locally and see what
happen.

Le ven. 21 juil. 2017 09:47, Mylène Bressan <notifications@github.com> a
écrit :

PR #409, opened 2017-07-24
Feature/backend/sanitize message body

sapiens-sapide:


src/backend/interfaces/REST/py.server/caliopen_api/message/message.py
```diff
@@ -68,17 +68,19 @@ def collection_post(self):
...
+        #     message = ObjectMessage(self.user.user_id, message_id=message_id)
+        #     try:
+        #         message.get_db()
+        #     except Exception as exc:
+        #         log.warn(exc)
+        #         raise ResourceNotFound
+        #
+        #     message.unmarshall_db()
+        #     return message.marshall_json_dict()
+        return Response(None, 301)
```
gdchamal: should return to correct v2 route not None

src/backend/main/go.backends/index/elasticsearch/index_lookup_test.go
```diff
@@ -10,7 +10,7 @@ func TestElasticSearchBackend_RecipientsSuggest(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	user_id := "f34528ca-b172-4df3-b27f-32cb18e1ad1f"
+	user_id := "8430b7e1-018e-482e-a06b-eb5e01a9ebcf"
```
gdchamal: it won't be easy to update the user_id any time we reimport. Testing strategy with volatile identifier have to be discussed better.

src/backend/interfaces/REST/py.server/caliopen_api/message/message.py
sapiens-sapide: Exception raised with commit #0dab108

src/backend/main/go.backends/index/elasticsearch/index_lookup_test.go
sapiens-sapide: Yes, definitely. Need to handle user_id & discussion_id with command line args (`go test -args`).

PR #422, opened 2017-07-28
Feature/backend/raw message

lisonferez:
Now we can fetch raw message.
But the file src/backend/protocols/go.imap_fetcher/cmd/caliopen_lmtpd/cli_cmds/serve.go is incomplete because, line 79, we need an user to fetch.
I don't tested if the method GetRemoteIdentity works. This method allows to fetch raw message because use information in RemoteIdentities of an user.


gdchamal:
Job will be done in another branch by @sapiens-sapide 
too many change since this PR. useful code will be copied into new branch

PR #436, opened 2017-08-21
add internalId to draft messages

iamdey:
+ allow compose new messages in multiple tabs
+ fix blink when new message is saved, the url remain the same, so it prevents re-rendering


iamdey:
if client is able de POST new message_id (using uuid), I'll be able to remove a lot of complexity:

* prepare & simplify offline capabilities
* only one url for compose & edit / reply (and so only one component & one redux flux instead of two)
* bug fix when editing same message on 2 different urls (see above)
* use only message_id instead of discussion_id / internal_id

/cc @gdchamal @sapiens-sapide

PR #439, opened 2017-08-24
Feature/backend/frontend can post draftID

sapiens-sapide:
by request from iamdey in PR #436

### DIFF
--- a/src/backend/main/py.main/caliopen_main/message/parameters/draft.py
+++ b/src/backend/main/py.main/caliopen_main/message/parameters/draft.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 import re
+import uuid
 from schematics.types import StringType
 from .message import NewMessage
 from caliopen_main.user.objects.identities import LocalIdentity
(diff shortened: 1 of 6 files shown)


src/backend/main/py.main/caliopen_main/message/parameters/draft.py
```diff
@@ -23,6 +25,15 @@ class Draft(NewMessage):
     body_html = ""
     body = StringType()
 
+    def validate_uuid(self, user_id):
```
iamdey: did it also validates, the message_id cannot be updated ?

sapiens-sapide: No. For now backend is still lazy:  there are lot of fields that should not be modified by user but are not checked when patching… [wip]

src/backend/main/py.main/caliopen_main/message/parameters/draft.py
```diff
@@ -23,6 +25,15 @@ class Draft(NewMessage):
...
+        if not self.message_id or not isinstance(self.message_id, uuid.UUID):
+            raise err.PatchUnprocessable(
+                message="missing or invalid message_id")
+        q = ModelMessage.objects.filter(user_id=user_id,
```
gdchamal: a get instead of a filter | first is better

Issue #440, opened 2017-08-24
dropdowns look ugly in safari/mac on contacts tab


sapiens-sapide:
![capture d ecran 2017-08-24 a 12 13 59](https://user-images.githubusercontent.com/1287627/29661961-f5f1cd72-88c5-11e7-93d6-9bd11fdd48ed.png)

myleneb:
This is because we don't use any vendor trick in css.
Something can be done with some little hack like in: https://codepen.io/fabiomcosta/pen/Ilzpa 
what do you think of it @iamdey ?

myleneb:
@dey 
I pushed a fix here: https://github.com/CaliOpen/Caliopen/compare/fix/frontend/SelectFieldGroup can you have a look and tell what you think? It works fine for chrome, firefox and webkit browser Epiphany (suppose it works too for opera) :)

iamdey:
it seems the good news is Caliopen is working on safari

PR #443, opened 2017-08-25
Fix/frontend/select field group

myleneb:
This fix issue https://github.com/CaliOpen/Caliopen/issues/440 "dropdowns look ugly in safari/mac on contacts tab"


src/frontend/web_application/src/components/form/SelectFieldGroup/style.scss
```diff
@@ -7,20 +7,61 @@
...
+      border-style: solid;
+      border-color: $co-color__fg__text transparent transparent;
+      pointer-events: none;
+    }
+  }
+
   &__select {
     @include o-form-element;
+
+    // sass-lint:disable-all
```
iamdey: As we discussed, it could be better to have this using a post css processor

iamdey:
Actually I don't like very much the "wrapper" idea: it sounds like it tries to fix something strange with a css trick.

First, can we get rid of this wrapper and apply styles directly to the select element ?

Afterward, what is the main purpose of this PR ?   
If I'm right we try to fix the default rendering of selects within different browser.
Imo this should be a "reset" styles.

Don't you think it can be done in https://github.com/CaliOpen/Caliopen/blob/master/src/frontend/web_application/src/styles/base/reset.scss ?

myleneb:
Wrapping `<select>` seems to be the only way to custom -webkit and other vendors styles.
As we can't add `:after` to `<select>`, we can't apply style directly to this element...

PR #448, opened 2017-08-30
Feature/backend/user setting again

gdchamal:
API to manage user settings
- GET /settings
- PATCH /settings

Settings are initialized when an user is created

### DIFF
--- a/src/backend/interfaces/REST/py.server/caliopen_api/user/settings.py
+++ b/src/backend/interfaces/REST/py.server/caliopen_api/user/settings.py
@@ -5,7 +5,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 import logging
 
 from pyramid.response import Response
-from caliopen_main.objects.settings import Settings as ObjectSettings
+from caliopen_main.user.objects.settings import Settings as ObjectSettings
 from cornice.resource import resource, view
 
 from ..base import Api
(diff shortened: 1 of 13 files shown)


src/backend/main/py.main/caliopen_main/user/core/user.py
```diff
@@ -334,6 +348,29 @@ def setup_system_tags(self):
             tag['date_insert'] = datetime.datetime.now(tz=pytz.utc)
             Tag.create(self, **tag)
 
+    def setup_settings(self):
+        """Create settings related to user."""
+        # XXX set correct values
```
sapiens-sapide: We will need to fill these defaults settings from config file.

src/backend/main/py.main/caliopen_main/user/objects/settings.py
```diff
@@ -32,5 +32,5 @@ class Settings(ObjectUser):
     }
 
     _model_class = ModelSettings
-    _pkey_name = 'device_id'
+    _pkey_name = None
```
sapiens-sapide: I don't understand why you are removing the _pkey_name attribute. «Settings» is an `ObjectUser` ; it belongs to an user and has `user_id` as its primary key.

src/backend/interfaces/REST/py.server/caliopen_api/user/settings.py
```diff
@@ -0,0 +1,57 @@
...
+    @view(renderer='json', permission='authenticated')
+    def get(self):
+        """Return user settings."""
+        objects = ObjectSettings.list_db(self.user.user_id)
+        settings = [x.marshall_dict() for x in objects]
+        return settings[0]
+
+    @view(renderer='json', permission='authenticated')
+    def patch(self):
+        """Update settings with payload.
```
sapiens-sapide: comments are copied from 'message'… ;-)

src/backend/main/py.main/caliopen_main/user/core/user.py
gdchamal: yes

PR #450, opened 2017-08-30
Install autoprefixer with postcss-loader

myleneb:
Install [postcss-loader](https://github.com/postcss/postcss-loader) and [Autprefixer](https://github.com/postcss/autoprefixer) to add CSS vendor prefixes.


src/frontend/web_application/webpack/config.js
```diff
@@ -51,7 +51,7 @@ const configureStylesheet = (config, filename = 'client_[name]', relativePath =
     },
     {
       test: /\.scss$/,
-      loader: extractTextPlugin.extract('style-loader', 'css-loader?sourceMap!sass-loader'),
+      loader: extractTextPlugin.extract('style-loader', 'css-loader?sourceMap!postcss-loader?sourceMap!sass-loader'),
```
pascalduez: Bear in mind that if `css-loader` `minimize` option is on, then `autoprefixer` will be run twice, because `minimize` uses `cssnano`. And I had a hard time making the cssnano version use our browsers config.

`autoprefixer` add prefixes, but also remove them if not needed. So you get the idea...

src/frontend/web_application/package.json
```diff
@@ -109,6 +109,7 @@
     "@kadira/react-storybook-addon-info": "^3.4.0",
     "@kadira/storybook": "^2.21.0",
     "@kadira/storybook-addon-knobs": "^1.7.1",
+    "autoprefixer": "^7.1.3",
```
iamdey: yarn.lock is missing

PR #458, opened 2017-09-01
Fix/frontend/new draft form

myleneb:
Make bottom buttons visible on mobile view (in NewDraftForm and ReplyForm).

### DIFF
--- a/src/frontend/web_application/src/components/NewDraftForm/style.scss
+++ b/src/frontend/web_application/src/components/NewDraftForm/style.scss
@@ -24,13 +24,14 @@
     border-left: 1px solid $co-color__bg__back;
   }
 
-  &__type {
-    @include flex-grid-size(shrink);
-  }
-
+  &__type { @include flex-grid-size(shrink); }
   &__type-icon { color: $co-color__fg__text; }
-
   &__type-label { display: none; }
+
+  &__date {
+    @include flex-grid-column(shrink);
+    margin-left: auto;
+  }
 
   &__author {
     @include flex-grid-column(shrink);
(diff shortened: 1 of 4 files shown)


src/frontend/web_application/src/components/NewDraftForm/style.scss
```diff
@@ -24,29 +24,24 @@
...
-
+  &__type { @include flex-grid-size(shrink); }
   &__type-icon { color: $co-color__fg__text; }
-
-  &__author {
-    @include flex-grid-column(shrink);
-    color: $co-color__fg__text--high;
-    font-weight: 600;
-  }
+  &__type-label { display: none; }
```
iamdey: it sounds `__type-label` is not overridden by a breakpoint.
So for what is it useful for ?

src/frontend/web_application/src/components/ReplyForm/style.scss
```diff
@@ -29,19 +29,18 @@
...
 
   &__type-icon { color: $co-color__fg__text; }
+  &__type-label { display: none; }
 
-  &__author {
-    @include flex-grid-column(shrink);
-    color: $co-color__fg__text--high;
-    font-weight: 600;
-  }
+  &__avatar-col { flex: none; }
```
iamdey: may be fix the dom instead ?

src/frontend/web_application/src/components/NewDraftForm/style.scss
myleneb: it is overridden by breakpoint(medium): `&__type-label { display: inline; }` :)

Issue #459, opened 2017-09-04
Responsive view doesn't display correctly


myleneb:
Between 380px and 640px large, search field and burger menu are missing:

![](https://framapic.org/yhGeN5llv8DS/9TWy207ddQGY.jpg)

myleneb:
close with https://github.com/CaliOpen/Caliopen/pull/460

PR #460, opened 2017-09-04
Fix responsive view for small screens

myleneb:
fix issue https://github.com/CaliOpen/Caliopen/issues/459: display responsive view consistently for screens < 640px

### DIFF
--- a/src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/presenter.jsx
+++ b/src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/presenter.jsx
@@ -56,12 +56,9 @@ class DiscussionItemActionsContainer extends Component {
     const { __ } = this.props;
 
     return (
+      // FIXME: DiscussionItemActionsContainer should swipe on mobile view
       <div
-        className={classnames(
-          'm-discussion-item-actions-container__actions',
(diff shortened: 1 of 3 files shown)


myleneb:
Also desactivate DiscussionItemActionsContainer on small screens

src/frontend/web_application/src/scenes/DiscussionList/components/DiscussionItemActionsContainer/presenter.jsx
```diff
@@ -57,7 +57,11 @@ class DiscussionItemActionsContainer extends Component {
 
     return (
       <div
-        className={classnames('m-discussion-item-actions-container__actions', { 'm-discussion-item-actions-container--active__actions': this.state.isActive })}
+        className={classnames(
+          'm-discussion-item-actions-container__actions',
+          { 'm-discussion-item-actions-container--active__actions': this.state.isActive },
+          'show-for-medium'
```
iamdey: I think we should completely disable `DiscussionItemActionsContainer` in parent components and open an issue for @Lauth0. 
We need a good solution that only works for all screen resolutions. What do think ?

myleneb: I finally restored it, and open issue https://github.com/CaliOpen/Caliopen/issues/464

PR #461, opened 2017-09-05
Fix/frontend/responsive profile connect api

iamdey:
TODO: set default value for im/email/addresse types (api does not support to not have this value)


iamdey:
I broke all the thing :( I close and do another

PR #465, opened 2017-09-06
Feature/backend/unfiltered messages list

sapiens-sapide:
To allow frontend to render the 'hammer' timeline, ie. unfiltered messages list.

### DIFF
--- a/src/backend/defs/go-objects/message.go
+++ b/src/backend/defs/go-objects/message.go
@@ -44,7 +44,7 @@ type Message struct {
 type MessagesListFilter struct {
 	Limit   int
 	Offset  int
-	Terms   map[string]string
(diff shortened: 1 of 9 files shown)


iamdey:
can you add `excerpt` prop (like in discussion) ?

sapiens-sapide:
excerpt is a discussion property. We do not have discussions object on this route.

iamdey:
excerpt is able to extract raw text from a mail to display it on the timeline. We still need it for the hammer.

If you can't extract it, it will be done client side (no idea how).

sapiens-sapide:
Following today's morning meeting, we'll modify API to return Message object with a new property : `excerpt`.

Issue #455, opened 2017-09-01
Message's plain body not correctly rendered.


sapiens-sapide:
Plain text bodies are rendered without `<pre>` tags, thus linefeeds are not rendered. 
[as discussed with @iamdey and @gdchamal, backend will return a new field : `body_is_plain` of type boolean to help frontend to correctly render the body].

sapiens-sapide:
backend feature proposed by PR #456

myleneb:
close with https://github.com/CaliOpen/Caliopen/pull/468

PR #468, opened 2017-09-11
fix/frontend/Message

myleneb:
Fix issue https://github.com/CaliOpen/Caliopen/issues/455
- Display pre or html depends on body_is_plain
- override html messages inline styles with default styles
- fix `<Message>` styles (avatar col and top bar)

### DIFF
--- a/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
+++ b/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
@@ -15,40 +15,6 @@ import './style.scss';
 
 const DropdownControl = withDropdownControl(Button);
 
-const MessageInfosContainer = ({ __, message, author, locale }) => {
-  const typeTranslations = {
-    email: __('message-list.message.protocol.email'),
-  };
-
-  return (
-    <div className="m-message__infos-container">
-      <TextBlock className="m-message__author">{author.address}</TextBlock>
-      {message.type &&
-        (<div className="m-message__type">
-          <span className="m-message__type-label">
-            {__('message-list.message.by', { type: typeTranslations[message.type] })}
-          </span>
-          {' '}
-          <Icon type={message.type} className="m-message__type-icon" spaced />
-        </div>
-      )}
-      <Moment className="m-message__date" format="LT" locale={locale}>
-        {message.date}
-      </Moment>
-    </div>
-  );
-};
-
-MessageInfosContainer.propTypes = {
-  author: PropTypes.shape({}).isRequired,
-  message: PropTypes.shape({}).isRequired,
-  locale: PropTypes.string,
-  __: PropTypes.func.isRequired,
-};
-MessageInfosContainer.defaultProps = {
-  locale: undefined,
-};
-
 class Message extends Component {
   static propTypes = {
     message: PropTypes.shape({}).isRequired,
@@ -117,6 +83,9 @@ class Message extends Component {
     );
 
     const dropdownId = uuidV1();
+    const typeTranslations = {
+      email: __('message-list.message.protocol.email'),
+    };
 
     return (
       <div className="m-message">
@@ -130,12 +99,20 @@ class Message extends Component {
           <div className={topBarClassName}>
             <MultidimensionalPi pi={message.pi} className="m-message__pi" mini />
 
-            <MessageInfosContainer
-              message={message}
-              author={author}
-              locale={locale}
-              __={__}
-            />
+            <TextBlock className="m-message__author">{author.address}</TextBlock>
+            {message.type &&
+              (<div className="m-message__type">
+                <span className="m-message__type-label">
+                  {__('message-list.message.by', { type: typeTranslations[message.type] })}
+                </span>
+                {' '}
+                <Icon type={message.type} className="m-message__type-icon" spaced />
+              </div>
+            )}
+
+            <Moment className="m-message__date" format="LT" locale={locale}>
+              {message.date}
+            </Moment>
 
             <DropdownControl toggle={dropdownId} className="m-message__actions-switcher">
               <Icon type="ellipsis-v" />
--- a/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
+++ b/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
@@ -13,7 +13,17 @@ import MessageActionsContainer from '../MessageActionsContainer';
 
 import './style.scss';
 
+const FOLD_HEIGHT = 80; // = .m-message__content--fold height
+
 const DropdownControl = withDropdownControl(Button);
+
+function generateStateFromProps(props) {
+  const { message } = props;
+
+  return {
+    isUnread: message.is_unread,
+  };
+}
 
 class Message extends Component {
   static propTypes = {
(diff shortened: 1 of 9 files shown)


src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -86,51 +48,44 @@ class Message extends Component {
...
   }
 
   render() {
     const { message, locale, onDelete, __ } = this.props;
     const author = message.participants.find(participant => participant.type === 'From');
     const subject = message.subject;
+
+    const bodyClassName = classnames(
+      'm-message__body',
+      { 'm-message__body--exerpt': this.state.isExerpt },
```
gdchamal: typo on exerpt (missing c) it's excerpt the correct word

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -171,21 +134,36 @@ class Message extends Component {
...
-          </div>
+          {subject &&
+            <TextBlock className="m-message__subject">
+              <Icon type="comments-o" rightSpaced />{subject}
+            </TextBlock>
+          }
+          {message.body_is_plain ? (
+            <pre
+              className={bodyClassName}
+              dangerouslySetInnerHTML={
```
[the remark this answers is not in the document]
iamdey: the conditions to when display excerpt belongs to an other User story.
Here is the draft of those conditions, feel free to arrange it: https://tree.taiga.io/project/caliopen-caliopen/task/439?kanban-status=815892

src/frontend/web_application/src/components/MessageList/components/Message/style.scss
```diff
@@ -1,107 +1,102 @@
...
-    padding-bottom: $co-component__spacing;
-    background: $co-color__fg__back;
+    padding: $co-component__spacing;
     overflow: hidden;
+    color: $co-color__fg__text;
 
-    &__content {
-      padding: $co-component__spacing 0;
-      color: $co-color__fg__text--low;
+    &--exerpt {
```
iamdey: excerpt should not contain any html elements nor line breaks, only plain text.

src/frontend/web_application/src/components/MessageList/components/Message/style.scss
```diff
@@ -1,107 +1,102 @@
...
+      &,
+      > * {
+        display: inline-block;
+      }
 
       &::after {
```
iamdey: excerpt should be a 200chars max length. 
the height of the excerpt div should be 1 line, so we can hide the rest and show an ellipsis when text is longer

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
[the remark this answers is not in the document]
myleneb: @iamdey : pour le moment `excerpt` c'est seulement un `substring()` sur le body, donc si on ne le `dangerouslySetInnerHTML` pas, ça affichera les balises HTML dans un simple `<div className="m-message__excerpt">{excerpt}</div>`

src/frontend/web_application/src/components/MessageList/components/Message/style.scss
sapiens-sapide: Excerpt could contain line break.

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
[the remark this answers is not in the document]
iamdey: pour récapituler:
* excerpt n'est utilisé que dans la timeline (pas de line breaks)
* pas de substring sur les body html (risque de casser le contenu du mail)
* `<pre>` sur body plain_text sans `dangerslySetInnerHtml`
* Il existe deux rendu de message fold/unfold le bouton toggle joue sur la hauteur de la div qui contient le body

Sur ce dernier point, se pose un nouveau pbm : les mails html avec beaucoup de vide en haut
-> faisons des ajustements lors de la phase alpha ?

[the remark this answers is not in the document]
sapiens-sapide: pour info, mon algo de génération de l'excerpt supprime les lignes vides des mails html

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -6,47 +6,24 @@ import ContactAvatarLetter from '../../../ContactAvatarLetter';
...
-  author: PropTypes.shape({}).isRequired,
-  message: PropTypes.shape({}).isRequired,
-  locale: PropTypes.string,
-  __: PropTypes.func.isRequired,
-};
-MessageInfosContainer.defaultProps = {
-  locale: undefined,
-};
+  return {
+    isUnread: message.is_unread,
```
iamdey: is_unread should not be in state: it is always a property of a message.

When an unread message is displayed the action set_read is sent. Then a middleware refresh the message and then you well see message.is_unread property up to date in props.

Finally, state should only be used to temporary set unsaved values (forms) or set a UI state depending on user interactions.

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -62,20 +39,19 @@ class Message extends Component {
...
+    isUnread: false,
+  };
+
+  componentWillMount() {
+    this.setState(prevState => generateStateFromProps(this.props, prevState));
   }
 
   componentDidMount() {
-    this.setBodyHeight();
+    this.setContentHeight();
```
iamdey: I didn't saw this but you might have a warning here, if I'm right.

You cannot setState in `componentDidMount` nor `render` methods because it produce a new render cycle. So in some cases it induces to an infinite loop.

For now I'm sure it is not a good practice unless setState is used in a setTimeout.

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -86,51 +62,78 @@ class Message extends Component {
...
-        body: body.substring(0, 140),
-      });
-    } else {
-      this.setState({
-        body,
-      });
-    }
-  }
+  setContentHeight = () => {
+    const bodyHeight = this.divElement.clientHeight;
```
iamdey: 👍

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -86,51 +62,78 @@ class Message extends Component {
...
+    };
+
     const topBarClassName = classnames(
       'm-message__top-bar',
-      { 'm-message__top-bar--active': this.state.isActive }
-    );
-    const bodyClassName = classnames(
-      'm-message__body__content',
-      { 'm-message__body__content--expanded': this.state.isExpanded }
+      { 'm-message__top-bar--is-unread': this.state.isUnread },
```
iamdey: I don't understand why this does not work `message.is_unread`

iamdey: btw the modifier could be on the Block: `m-message--is-unread__top-bar` (same for the next class: `m-message--is-unread__subject`)

src/frontend/web_application/src/components/MessageList/components/Message/style.scss
```diff
@@ -80,28 +52,70 @@
...
+      color: $co-color__fg__text;
     }
+  }
 
-    &__avatar {
-      @include m-avatar--size(small);
-      display: block;
+  &__content {
+    position: relative;
+    transition-duration: .15s; // transition-duration is overrided by JS
```
iamdey: overridden

src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
iamdey: ok if you can't deal with in this PR (but we will have to fix it soon)

PR #470, opened 2017-09-12
"Forgot password" link and component UI

myleneb:
This creates a component `<PasswordResetForm>` displayed in `<ResetPassword>` scene (in `<Auth>` with path "/auth/reset-password") and adds a "Forgot password?" link on `<SigninForm>`

By now it only asks for `username` and (TODO:) will process form to send an email with instructions (or whatever will be the reset password process, depending on UX).

Captures: https://tree.taiga.io/project/caliopen-caliopen/task/442


src/frontend/web_application/src/components/form/FormGrid/style.scss
```diff
@@ -31,6 +31,10 @@
 
       padding-right: map_get($co-form__spacing, 'medium');
 
+      &--full {
+        padding-right: 0;
```
iamdey: If I look only at this directive, it's sounds very weird: a right padding doesn't make a column full width itself.

If I understand right the `--full` modifier is used to fix an incorrect padding added at medium breakpoint. 
What if we just remove the padding above ?

src/frontend/web_application/src/scenes/ResetPassword/presenter.spec.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Presenter from './presenter';
+
+describe('scene - ResetPassword', () => {
```
iamdey: 👍

src/frontend/web_application/src/styles/object/o-clickable.scss
```diff
@@ -77,9 +77,16 @@
 }
 
 @mixin o-clickable--inline {
-  margin-left: 1px;
-  background: $co-color__fg__back--lower;
-  font-size: $co-font__size--xsmall;
+  display: inline;
+  padding: 0;
```
iamdey: why those reset ?

may be `reset.scss` instead ?

src/frontend/web_application/src/components/form/FormGrid/style.scss
myleneb: ok to remove "padding-right" but so, we have to add a modifier "--right-spaced" for columns that needs this right padding (most of time they need it).

src/frontend/web_application/src/styles/object/o-clickable.scss
myleneb: mmm I don't think reset.scss is the right place.
The problem here is that `Button` has its own styles (brought by `o-clickable--button`, including padding) and so its modifiers inherit these styles. Modifier `--inline` should not contain any padding nor line-height, etc... so this is the simplest way to override styles only for this modifier.

src/frontend/web_application/src/components/form/FormGrid/style.scss
myleneb: btw would it be good to set an `rightSpaced` default prop to `true`?

```
FormColumn.defaultProps = {
  className: undefined,
  size: undefined,
  fluid: false,
  bottomSpace: false,
  rightSpaced: true,
};
```

src/frontend/web_application/src/components/form/FormGrid/index.jsx
```diff
@@ -22,13 +23,15 @@ FormColumn.propTypes = {
...
   fluid: PropTypes.bool,
+  rightSpace: PropTypes.bool,
   bottomSpace: PropTypes.bool,
 };
 FormColumn.defaultProps = {
   className: undefined,
   size: undefined,
   fluid: false,
   bottomSpace: false,
+  rightSpace: true, // --right-space style is default for FormColumn
```
iamdey: 👍

src/frontend/web_application/src/styles/object/o-clickable.scss
iamdey: ok so, can I propose you to add the modifier `--inline-block` with `@include o-clickable--button;` to the Button (which will be the default display mode) (Careful to display=expanded which could need it too) ?

src/frontend/web_application/src/components/form/FormGrid/style.scss
iamdey: I forgot to answer "yes"

src/frontend/web_application/src/components/Button/index.jsx
```diff
@@ -35,8 +37,9 @@ const Button = ({ children, className, icon, display, color, shape, responsive,
         'm-button--secondary': color === 'secondary',
         'm-button--success': color === 'success',
 
-        'm-button--expanded': display === 'expanded',
```
iamdey: It makes sense to me to keep expanded in display modes and duplicates css directives (it's not coherent to have `expanded display="inline"` for example):

```
&--inline-block {
  @include o-clickable--button;
}

&--expanded {
  @include o-clickable--button;
  @include o-clickable--expanded;
  width: 100%;  // useful ?
}
```

src/frontend/web_application/src/components/Button/index.jsx
```diff
@@ -78,7 +82,8 @@ Button.defaultProps = {
   children: null,
   shape: null,
   icon: null,
-  display: null,
+  expanded: false,
+  display: 'inline-block', // default style for button is --inline-block
```
iamdey: yeah, I think this is really better

PR #472, opened 2017-09-15
Feature/backend/settings part2

gdchamal:
Adapt settings to new version as defined in US 363.

### DIFF
--- a/src/backend/defs/rest-api/objects/Settings.yaml
+++ b/src/backend/defs/rest-api/objects/Settings.yaml
@@ -3,7 +3,7 @@ type: object
 properties:
   default_locale:
     type: string
-    default: fr_FR
+    default: fr-FR
   message_display_format:
     type: string
     default: rich_text
--- a/src/backend/defs/rest-api/objects/Settings.yaml
+++ b/src/backend/defs/rest-api/objects/Settings.yaml
@@ -3,15 +3,25 @@ type: object
 properties:
   default_locale:
     type: string
+    default: fr_FR
   message_display_format:
     type: string
+    default: rich_text
   contact_display_order:
     type: string
+    default: given_name
+  contact_display_format:
+    type: string
+    default: family_name, given_name
   notification_enabled:
     type: boolean
+    default: true
   notification_message_preview:
     type: string
+    default: always
   notification_sound_enabled:
     type: boolean
+    default: false
   notification_delay_disappear:
     type: integer
+    default: 10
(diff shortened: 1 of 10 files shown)


src/backend/main/py.main/caliopen_main/user/parameters/settings.py
```diff
@@ -3,19 +3,29 @@
...
 from schematics.models import Model
-from schematics.types import StringType, IntType
+from schematics.types import StringType, IntType, BooleanType
+
+MESSAGE_FORMAT_CHOICES = ['rich_text', 'plain_text']
+CONTACT_FORMAT_CHOICES = ['given_name', 'family_name',
+                          'family_name, given_name']
+CONTACT_ORDER_CHOICES = ['family_name', 'given_name']
+PREVIEW_CHOICES = ['off', 'always']
+DELAY_CHOICES = [5, 10, 10, 30]
```
sapiens-sapide: 5, 10, **20**, 30 may be ?

src/backend/main/py.main/caliopen_main/user/parameters/user.py
```diff
@@ -27,6 +28,7 @@ class NewUser(Model):
     params = DictType(StringType())
     password = StringType(required=True)
     recovery_email = EmailType(required=True)
+    settings = ModelType(Settings, default=lambda: {})
```
sapiens-sapide: `default` could be a func that returns default settings from config file ?

gdchamal: default value are a mess and not handled correctly for the moment. default values are defined both in schematics Settings and in swagger object.

I think that I'll remove default value from schematics later and try to see how to handle only API side, ideally using a configuration file.

src/backend/main/py.main/caliopen_main/user/parameters/settings.py
gdchamal: I followed the spec ;)

src/backend/defs/rest-api/objects/Settings.yaml
```diff
@@ -1,19 +1,27 @@
...
 type: object
 properties:
-  default_language:
-    type: string
-  default_timezone:
-    type: string
-  date_format:
+  default_locale:
     type: string
+    default: fr_FR
```
iamdey: I was wrong, it should `fr-FR`, (this what I get from browser)

gdchamal: fixed in 8ab87b3879ea6a48d62968500bafd81f50d77915

PR #473, opened 2017-09-15
apply i18n from settings, separate locale & language

iamdey:
+ add I18nProvider & SettingsProvider
+ add PageNotFound route & Component


iamdey:
there's a non blocking bug when applying an other language

* first re-render: the interface is partially translated, options in fields are not translated
* when refreshing: the whole form content stays in the default language
* All is going well when coming from an other page

Issue #464, opened 2017-09-06
DiscussionItemActionsContainer UI on mobile view


myleneb:
on mobile view, `DiscussionItemActionsContainer` is hiding `DiscussionItem`

![](https://camo.githubusercontent.com/2d6a4bb26bc1df93c64d7090263ef0d688afd6cd/68747470733a2f2f6672616d617069632e6f72672f6e704d6e636e6644777156552f7773354131575563714444532e6a7067)

as discussed with @Lauth0 it should swipe.
What do you think @iamdey ?
poke @judbd

iamdey:
the swipe can be done using the library already in the project https://www.npmjs.com/package/react-tappable

myleneb:
close with https://github.com/CaliOpen/Caliopen/pull/492

PR #484, opened 2017-09-20
Feature/frontend/swipe

myleneb:
This creates a simple Swipe component and adds it into DiscussionItemContainer
(issue https://github.com/CaliOpen/Caliopen/issues/464)


iamdey:
closed ?

I've tried the branch, it was almost good, there are just some adjustments to make it more fluid/natural.

The strange thing is that the animation starts when swipe is released instead of following the finger.

PR #485, opened 2017-09-20
Make cassandra protocol_version configurable for python services

gdchamal:
Using scylladb we must use protocol_version = 3 to avoid some warning (and automatic downgrade of protocol by python driver).

This pull request permit to define the protocol_version in yaml configuration file


sapiens-sapide:
Is it still OK with cassandra ? (I'm using cassandra on my Mac)

gdchamal:
Sound OK with a cassandra 3.x serie and protocol_version = 3:

https://datastax.github.io/python-driver/api/cassandra.html#cassandra.ProtocolVersion

PR #517, opened 2017-10-09
Refacto Dropdown

myleneb:
- update components using Dropdown
- replace DropdowMenu with isMenu props on Dropdown
- remove foundation's  Dropdown export

### DIFF
--- a/src/frontend/web_application/src/components/Dropdown/index.jsx
+++ b/src/frontend/web_application/src/components/Dropdown/index.jsx
@@ -121,7 +121,6 @@ class Dropdown extends Component {
       // otherwise return prevState.offset
       const newOffset = prevState.isOpen ? prevState.offset : this.updateOffset();
 
-      // otherwise return prevState.offset
       if (isVisible !== prevState.isOpen) { this.props.onToggle(isVisible); }
 
       return {
@@ -136,7 +135,7 @@ class Dropdown extends Component {
     const control = this.dropdownControl;
     const dropdown = this.dropdown;
 
-    // if no dropdownControl decared, return empty offset
+    // if no dropdownControl declared, return empty offset
     // otherwise, return new offset
     return control ? getOffset(alignRight, position, control, dropdown) : {};
   }
--- a/src/frontend/web_application/src/components/Dropdown/index.jsx
+++ b/src/frontend/web_application/src/components/Dropdown/index.jsx
@@ -46,7 +46,11 @@ class Dropdown extends Component {
     children: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
     className: PropTypes.string,
     closeOnClick: PropTypes.bool, // should Dropdown close on click?
-    closeOnClickExceptSelectors: PropTypes.arrayOf(PropTypes.shape({})),
+    // closeOnClickExceptRefs: array of refs that should not close dropdown on click
+    // usage:
+    // <div ref={foo => this.foo =foo} />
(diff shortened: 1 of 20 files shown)


src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -1,145 +1,172 @@
...
-    if (this.onToggle) {
-      this.$dropdown.on('show.zf.dropdown', () => {
-        this.onToggle(true);
-      });
-      this.$dropdown.on('hide.zf.dropdown', () => {
-        this.onToggle(false);
-      });
-    }
+  componentDidMount() {
+    this.dropdownControl = document.getElementById(`${CONTROL_PREFIX}-${this.props.id}`);
```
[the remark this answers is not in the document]
iamdey: We can change this later because it requires an other structure (with some HoC maybe).

src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -1,145 +1,171 @@
...
+      const exceptSelectors = this.props.closeOnClickExceptSelectors;
 
-        if (this.$dropdown.is(target) || this.$dropdown.find(target).length) {
-          return;
-        }
+      const dropdownClick = !this.props.closeOnClick &&
+        (this.dropdown === target || this.dropdown.contains(target));
 
+      const exeptSelectorsClick = exceptSelectors &&
+        exceptSelectors.find(selector => (selector === target));
```
iamdey: I'm not sure this works : target is not a string or may be selector is not a cssSelector but something more complicated.
(it looks like something a bit magic)

src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -1,145 +1,171 @@
...
+    this.toggle(this.props.show);
+    document.addEventListener('click', this.handleDocumentClick);
+    if (this.props.closeOnScroll) { window.addEventListener('scroll', this.handleDocumentScroll); }
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.show !== nextProps.show) {
-      this.show(nextProps.show);
-    }
+    if (this.props.show !== nextProps.show) { this.toggle(nextProps.show); }
```
iamdey: _(my preference goes for multiple lines for conditions, that I find more readable. Also you might want to drop braces to keep one-line condition, but I still find this less readable than multiple lines + braces)_

_(less lines of code is not better, less code is :) )_

Do you really prefer one-line condition ? We can set it in eslint.

src/frontend/web_application/src/components/Dropdown/style.scss
```diff
@@ -1,12 +1,25 @@
 @import '../../styles/common';
+@import '../../styles/object/o-triangle';
+@import '../../styles/vendor/bootstrap_foundation-sites';
```
iamdey: I don't understand why do you need foundation here

src/frontend/web_application/src/styles/vendor/bootstrap_foundation-sites.scss
```diff
@@ -19,7 +19,7 @@
 @include foundation-menu;
 @include foundation-menu-icon;
 // @include foundation-drilldown-menu;
-@include foundation-dropdown;
+// @include foundation-dropdown;
```
iamdey: 🥇

PR #523, opened 2017-10-12
Load configuration from file & overrides with env var

iamdey:
TODO:

* [x] support load config from file set by cli args
* [x] load config in frontend
* [x] fix signout > signin redirect


src/frontend/web_application/config/server.default.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  version: process.env.CALIOPEN_VERSION || 'master',
```
gdchamal: Version is not be a configuration parameter, but a variable defined during build.

Actually the package.json is updated thru devtools/make_package command

iamdey: right, but I keep it configurable (for example for a specific branch build). So default will read from package.json

PR #524, opened 2017-10-12
We can define a configuration file to whitelist allowed user recovery…

gdchamal:
… email for registration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 - API to allow users to change their password, with email notification
+- Compute first contact privacy features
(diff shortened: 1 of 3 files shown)


gdchamal:
That a first whitelist strategy a minima.
If we need to add random it will be easy

For user feedback when it fail, the API raise a 401 with an error message that email is not in whitelist. You want another HTTP code ?

iamdey:
we need some frontend adjustments, the feedback from api is not displayed as is in the frontend.

sapiens-sapide:
This code allows an user to create more than one account with the same `recovery_email`. Is it ok ?

gdchamal:
For frontend impact, I think we can only display a disclosure message that you must have a registered email (display a link to alpha subscribe form)

About multiple account with the same recovery email, I think it's not a problem

PR #525, opened 2017-10-13
Feature/frontend/reply

iamdey:
handle reply

it displays the excerpt of the parent if the parent is not the last one

capture d ecran de 2017-10-13 18-06-30


iamdey:
* [ ] something is wrong reply from timeline because draft is not yet well initialized

PR #556, opened 2017-10-25
Upgrade react-redux-scroll

josepot:
I think that with the [latest version of `react-redux-scroll`](https://github.com/josepot/react-redux-scroll/pull/8/files) there won't be a need to check for the `BUILD_TARGET` value.

See the discussion [here](https://github.com/josepot/react-redux-scroll/pull/7#issuecomment-339272974).

cc @iamdey

### DIFF
--- a/src/frontend/web_application/package.json
+++ b/src/frontend/web_application/package.json
@@ -83,7 +83,7 @@
     "react-moment": "^0.6.5",
     "react-redux": "^5.0.6",
     "react-redux-notify": "^2.0.0",
-    "react-redux-scroll": "^0.0.10",
+    "react-redux-scroll": "^0.0.12",
     "react-router-dom": "^4.2.2",
     "react-router-redux": "next",
     "react-swipeable": "^4.1.0",
--- a/src/frontend/web_application/src/scenes/MessageList/components/DraftForm/index.js
+++ b/src/frontend/web_application/src/scenes/MessageList/components/DraftForm/index.js
@@ -1,16 +1,11 @@
 import { createSelector } from 'reselect';
 import { bindActionCreators, compose } from 'redux';
 import { connect } from 'react-redux';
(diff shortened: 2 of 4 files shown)


iamdey:
I've some trouble to see the result of travis (I'm in a train), but I think this is due to `IS_SSR` that must be added to build scripts in `package.json` and `webpack/config.js` (`congifEnv`) (I prefer not change `HAS_SSR` because it's a dev option to say if express server has ssr or not and is always true in prod env.)

Also we are in high alpha release stabilization this week. So I will not merge new PR in master until 2 weeks (at least). Do you mind to create a new PR using develop base branch ?

Big thanks to you to make this PR. <3

josepot:
> I've some trouble to see the result of travis (I'm in a train), but I think this is due to IS_SSR that must be added to build scripts in package.json and webpack/config.js (congifEnv) (I prefer not change HAS_SSR because it's a dev option to say if express server has ssr or not and is always true in prod env.)

Actually travis is crashing for a completely unrelated thing: a linting error that has nothing to do with my changes: 

```
/home/travis/build/CaliOpen/Caliopen/src/frontend/web_application/src/layouts/Page/components/Header/presenter.jsx
 7:8 error 'SearchField' is defined but never used no-unused-vars
```

There is no need for you to setup the value of `IS_SSR` to `true`... That's only necessary if you want to completely disable the library. With the latest changes that won't be necessary because nothing that doesn't come from `componentWillMount` will ever access the `window` object. So, setting that env variable may give you a tiny performance bust on the server side, but that's it.

> Also we are in high alpha release stabilization this week. So I will not merge new PR in master until 2 weeks (at least). Do you mind to create a new PR using develop base branch ?

No worries! I totally get it. I will close this PR and I will re-open it against your `develop` branch. Don't feel like you have to merge it, no ofense taken if you prefer not to.

PR #572, opened 2017-10-27
Hot fix call-to-action

myleneb:
- no more principalActionSelector: all actions are stored in
availableActionsSelector
- display same call-to-action on every page (ComposeButton,
CreateContactButton)
- hide ComposeContactButton and ReplyButton


iamdey:
\o/

PR #623, opened 2017-11-08
Connect PasswordFom 

myleneb:
Returns notifications for success and errors.


src/frontend/web_application/src/scenes/UserSecurity/components/PasswordDetails/index.jsx
```diff
@@ -6,38 +6,60 @@ import PasswordForm from '../PasswordForm';
 import TextBlock from '../../../../components/TextBlock';
 import './style.scss';
 
+function generateStateFromProps(props, prevState) {
+  return {
+    ...prevState,
+    editMode: prevState.editMode && props.updated && false,
```
iamdey: is this equivalent to `prevState.editMode && !props.updated` ?

src/frontend/web_application/src/scenes/UserSecurity/presenter.jsx
```diff
@@ -11,40 +12,76 @@ class UserSecurity extends Component {
...
-    user: undefined,
+  state = {
+    updated: false,
   };
 
   componentDidMount() {
     this.props.requestUser();
   }
 
+  handleSubmit = (data) => {
```
iamdey: we will have to move this to redux for consistency & separation of concerns.

For now it sounds to work so ok

PR #624, opened 2017-11-08
SearchResults styles

myleneb:


src/frontend/web_application/src/scenes/SearchResults/style.scss
```diff
@@ -0,0 +1,29 @@
...
+
+    &:hover {
+      color: $co-color__primary--high;
+    }
+  }
+
+  &__item {
+    background: $co-color__fg__back;
+
+    // overrides 's-message-item' style in component/MessageItem/style.scss
```
iamdey: that's not ITCSS/BEM :(

no way to make it better ?

I didn't fixed the probable name conflict that can happen in Timeline's MessageItem and Search one.

May be do this in Search's MessageItem and check if this is a problem first ?

myleneb: I know this is not good :( 
but it's the only way I found to add styles on search `MessageItem` without impacting `Timeline` styles.
The cleanest solution would be a one (and only one) `MessageItem` component which would be used in both search and Timeline. what do you think of that?

iamdey: I have 3 solutions in minds so: 

* yours  : refactor MessageItem (but be careful because it might add complexity or lack of flexibility)
* change the name of Search/MessageItem (so it is a brand new component and then try to reused components sub-component)
* configure unique css prefix or suffix unique classnames to prevent conflicts at all (but it requires to play again with webpack)

myleneb: ok for changing name of component. what do you think of `MessageResultItem` ?

PR #634, opened 2017-11-10
Feature/backend/reset password

sapiens-sapide:
Implements reset password API, as describe in https://tree.taiga.io/project/caliopen-caliopen/task/416.
**NB** : 
1. A new lookup table — `user_recovery_email` — has to be populated into Cassandra store. Logic is in the code of this PR (for caliopen setup cli and for new user creation), but it is still required to build the table for users already created into alpha.
2. The file `configs/caliopne-go-api_dev.yaml` has been modified to embed notification related configuration.
3. A new directory — `src/defs/notifiers/templates` — must be copied to api2 instances, with templates files within it.


src/backend/interfaces/REST/go.server/operations/users/user.go
```diff
@@ -57,6 +58,97 @@ func PatchUser(ctx *gin.Context) {
...
+	err := ctx.BindJSON(&payload)
+	if err != nil {
+		e := swgErr.New(http.StatusUnprocessableEntity, err.Error())
+		http_middleware.ServeError(ctx.Writer, ctx.Request, e)
+		ctx.Abort()
+		return
+	}
+
+	if payload.RecoveryMail == "" && payload.Username == "" {
+		e := swgErr.New(http.StatusUnprocessableEntity, "neither username nor recovery email provided, at least one required")
```
gdchamal: it's a Bad request (400) not an unprocessable entity (422) HTTP error code to return

src/backend/interfaces/REST/go.server/operations/users/user.go
```diff
@@ -57,6 +58,97 @@ func PatchUser(ctx *gin.Context) {
...
+	return
+}
+
+// ResetPassword handles POST on /passwords/reset/:reset_token
+// payload should be a json with new password
+func ResetPassword(ctx *gin.Context) {
+
+	token := ctx.Param("reset_token")
+
+	if token == "" {
```
gdchamal: Bad request HTTP code

PR #643, opened 2017-11-15
Feature/backend/contact update worker

gdchamal:
Add a nats handler to process contact_update message and compute PI for this contact.

### DIFF
--- a/src/backend/interfaces/NATS/py.client/caliopen_nats/subscribers.py
+++ b/src/backend/interfaces/NATS/py.client/caliopen_nats/subscribers.py
@@ -5,8 +5,11 @@ from __future__ import absolute_import, print_function, unicode_literals
 import logging
 import json
 
-from caliopen_nats.delivery import UserMessageDelivery
 from caliopen_main.user.core import User
+from caliopen_main.contact.objects import Contact
+
+from caliopen_nats.delivery import UserMessageDelivery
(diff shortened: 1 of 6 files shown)


src/backend/interfaces/NATS/py.client/caliopen_nats/subscribers.py
```diff
@@ -55,10 +58,23 @@ def handler(self, msg):
...
     """Handler for contact action message."""
 
+    def process_update(self, msg, payload):
+        """Process a contact update message."""
+        # XXX validate payload structure
+        if 'user_id' not in payload or 'contact_id' not in payload:
+            raise Exception('Invalid contact_update structure')
+        user = User.get(payload['user_id'])
+        contact = Contact(user.user_id, contact_id=payload['contact_id'])
+        contact.get_db()
```
sapiens-sapide: do not forget to `contact.unmarshall_db()` before processing, do you ?

sapiens-sapide: we could simply add `methods retrieve_db()` and `retrieve_index()` that do `get_xxx()` then `unmarshall_xxx()` under the hood.

gdchamal: that's another story, we will define such interfaces later

PR #644, opened 2017-11-15
Fix/frontend/import contact file size

### LINKED ISSUE #592: Define maximal size for contact import
Have to find the technical limit if easy, else define a know working limit.

iamdey:
fix #592


gdchamal:
nginx configuration updated on client and api server to match this 5M limit:

https://github.com/CaliOpen/deploy-alpha/pull/17

Issue #642, opened 2017-11-15
Change password not correctly implemented


sapiens-sapide:
Trying to change password (route `/user/security`) in alpha's UI fails with error « l'ancien mot de passe ne correspond pas ».  
Under the hood, frontend sends an empty PATCH to API : 

```
{"current_state":{}}
```

sapiens-sapide:
[update] : to reproduce in UI : 
* go to user's account page
* clic « change » password button
* enter your old password, then, your old password again two times (meaning you do not change the password)

=> frontend tries to send an empty PATCH

PR #646, opened 2017-11-15
Fix handleSubmit in PasswordForm

myleneb:
fix https://github.com/CaliOpen/Caliopen/issues/642


src/frontend/web_application/src/scenes/UserSecurity/components/PasswordForm/index.jsx
```diff
@@ -80,7 +79,13 @@ class PasswordForm extends Component {
     const { formValues } = this.state;
     const user = { password: formValues.newPassword };
     const original = { password: formValues.password };
-    const data = calcObjectForPatch(user, original);
+
+    const data = {
```
iamdey: or:

```
const data = {
   password: formValues.newPassword,
   current_state: {
      password: formValues.password,
   },
};
```

PR #647, opened 2017-11-15
Fix UserSecurity

myleneb:
When you tried to access user/security directly via browser bar, you got an error.
this fix this bug.


iamdey:
There's probably something to add to the changelog ?

PR #663, opened 2017-11-22
Refactor contacts-middleware

myleneb:
Update `ContactBook` when you remove or create a contact.
Also add spinner button on `Contact`.
While testing it appears that if you delete user's contact it breaks app so I also disable delete button on user's contact.


iamdey:
done in #680

Issue #671, opened 2017-11-23
sort messages


iamdey:
messages are sorted by date_insert
It's normal behavior for message in reception but not for messages submitted because it's the date when the draft has been created.

myleneb:
Fixed by https://github.com/CaliOpen/Caliopen/pull/679
Also see: https://github.com/CaliOpen/Caliopen/issues/684

PR #679, opened 2017-11-28
Fix/frontend/sort messages

myleneb:
Fix https://github.com/CaliOpen/Caliopen/issues/671

This checks if `message` is from `user` and if so, sorts and fills message with `message.date` instead of `message.date_insert` both in `Timeline` and `MessageList`.

### DIFF
--- a/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
+++ b/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
@@ -74,6 +74,23 @@ class Message extends Component {
     }));
   }
 
+  renderDate = () => {
+    const { message, settings: { default_locale: locale }, isMessageFromUser } = this.props;
+    const hasDate = (isMessageFromUser && message.date)
+      || (!isMessageFromUser && message.date_insert);
+
+    return (
+      <div className="m-message__date">
+        {hasDate &&
+          <Moment format="LT" locale={locale}>
+            {isMessageFromUser ? message.date : message.date_insert}
+          </Moment>
+        }
+      </div>
+
+    );
+  }
+
   renderMessageContent = () => {
     const { message } = this.props;
 
@@ -109,8 +126,8 @@ class Message extends Component {
 
   render() {
     const {
-      message, settings: { default_locale: locale }, onDelete, onMessageUnread, onMessageRead,
-      onReply, onCopyTo, onEditTags, __, isMessageFromUser,
+      message, onDelete, onMessageUnread, onMessageRead,
+      onReply, onCopyTo, onEditTags, __,
     } = this.props;
     const author = getAuthor(message);
     const typeTranslations = {
@@ -149,10 +166,7 @@ class Message extends Component {
                 <Icon type={message.type} className="m-message__type-icon" spaced />
               </div>
             )}
-            {message.date_insert &&
-              <Moment className="m-message__date" format="LT" locale={locale}>
-                {isMessageFromUser ? message.date : message.date_insert}
-              </Moment> }
+            {this.renderDate()}
 
             <DropdownControl toggleId={this.dropdownId} className="m-message__actions-switcher" icon="ellipsis-v" />
 
--- a/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
+++ b/src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
@@ -11,6 +11,8 @@ import TextBlock from '../../../TextBlock';
 import MultidimensionalPi from '../../../MultidimensionalPi';
 import Dropdown, { withDropdownControl } from '../../../../components/Dropdown';
 import MessageActionsContainer from '../MessageActionsContainer';
+import { getAuthor } from '../../../../services/message';
+
 import './style.scss';
 
 const DropdownControl = withDropdownControl(Button);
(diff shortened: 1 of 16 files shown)


src/frontend/web_application/src/components/MessageList/components/Message/presenter.jsx
```diff
@@ -147,7 +151,7 @@ class Message extends Component {
             )}
             {message.date_insert &&
```
iamdey: the condition here is wrong when `isMessageFromUser` is true (and can broke the app if date is undefined). May be this can be done in a `renderDate` method.

src/frontend/web_application/src/components/MessageList/index.js
```diff
@@ -1,7 +1,19 @@
...
+import { connect } from 'react-redux';
 import { withTranslator } from '@gandi/react-translate';
 import Presenter from './presenter';
 
+const userSelector = state => state.user;
+
+const mapStateToProps = createSelector(
+  [userSelector],
+  usertState => ({
+    user: usertState.user,
```
iamdey: careful, userState might be undefined when starting the appllication (the request not yet fully loaded)

And an HoC can be useful : e.g withUser()

src/frontend/web_application/src/scenes/Timeline/index.js
```diff
@@ -7,14 +7,17 @@ import { requestMessages, loadMore, hasMore } from '../../store/modules/message'
...
-  [timelineSelector, messagesSelector],
-  (timeline, messagesById) => ({
+  [timelineSelector, messagesSelector, userSelector],
+  (timeline, messagesById, userState) => ({
     messages: timeline && timeline.messages.map(id => messagesById[id])
       .sort((a, b) => new Date(b.date_insert) - new Date(a.date_insert)),
     hasMore: timeline && hasMore(timeline),
     isFetching: timeline && timeline.isFetching,
     didInvalidate: timeline && timeline.didInvalidate,
+    user: userState.user,
```
iamdey: userState undefined ?

src/frontend/web_application/src/services/message/index.js
```diff
@@ -9,3 +9,13 @@ export const sortMessages = (messages, reversed) => messages.sort((a, b) => {
 export const renderParticipant = participant => `${participant.label} (${participant.address})`;
 
 export const getLastMessage = messages => sortMessages(messages, true)[0];
+
+export const getAuthor = message => message.participants && message.participants.find(participant => participant.type === 'From');
+
+export const isMessageFromUser = (message, user) => {
+  const author = getAuthor(message);
+  const authorId = author && author.contact_ids[0];
```
iamdey: in case of an unexpected reason there's more than one contact_id, it may not work correctly.

Also the name of the variable is not self-explanatory:

authorId -> authorContactId
userId -> userContactId

Imo, you should use `author.contact_ids.some(......)` to compare the userContactId.

src/frontend/web_application/src/services/message/index.js
```diff
@@ -9,3 +9,13 @@ export const sortMessages = (messages, reversed) => messages.sort((a, b) => {
...
+  const userId = user && user.contact.contact_id;
+
+  return userId && authorId && userId === authorId;
```
iamdey: I think userId && authorId MUST be set, it sounds not useful to do a condition with, isn't it?

In case those are not set this probably is a bug.

iamdey: btw, a unit test would be great ;)

src/frontend/web_application/src/hoc/user.js
```diff
@@ -0,0 +1,15 @@
+import { createSelector } from 'reselect';
+import { connect } from 'react-redux';
+import { UserSelector } from '../store/selectors/user';
+
+const mapStateToProps = createSelector(
+  [UserSelector],
+  userState => ({
+    user: userState.user,
+    isFetching: userState.isFetching,
```
iamdey: this will be a problem (it inherent to HoC generally):

if we use two HoC that provide the same property, it will be overwritten by the last one if I'm correct. 

It would be better to make a self-explanatory property like `isUserFetching`

Currently the best way to prevent this kind of error is to use a Render Prop instead of HoC cf. https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce

src/frontend/web_application/src/store/selectors/user.js
```diff
@@ -0,0 +1 @@
+export const UserSelector = state => state.user;
```
iamdey: I prefer keep `UpperCamelCase` format for class and simple `camelCase` for functions

src/frontend/web_application/src/scenes/UserPrivacy/index.jsx
```diff
@@ -1,25 +1,16 @@
...
-);
-
 const mapDispatchToProps = dispatch => bindActionCreators({
   requestUser,
 }, dispatch);
 
 export default compose(
-  connect(mapStateToProps, mapDispatchToProps),
+  connect(null, mapDispatchToProps),
+  withUser(true),
```
iamdey: unused `true`

src/frontend/web_application/src/scenes/UserPrivacy/index.jsx
```diff
@@ -1,25 +1,16 @@
...
 import { requestUser } from '../../store/modules/user';
 import Presenter from './presenter';
 
-const userSelector = state => state.user;
-
-const mapStateToProps = createSelector(
-  [userSelector],
-  userState => ({
-    user: userState.user,
-    isFetching: userState.isFetching,
```
iamdey: keep this line

Issue #673, opened 2017-11-24
Reset password will not work if a recovery email is set for multiple account.


sapiens-sapide:
Current backend implementation of the reset password procedure will not handle the case where the recovery email entered in the form leads to find more than one user.

gdchamal:
In fact only one recovery email can exist at a time. As we have a recovery_email table that can contain uniaue recovery email.

We must restrict to an unique recovery_email or remove logic around lookup up for such value and capability to reset password with such value.

sapiens-sapide:
Must discuss it ! 
* Option 1 : unique recovery_email, meaning that one can not create multiple accounts with the same recovery_email
* Option 2 : allow to reuse the same recovery_email for different accounts : means that the reset password process could only be triggered with username : ie user who wants to reset password MUST remember its username.

LaurentChemla:
option 1

iamdey:
option 1 (comme le chef a dit)

but I don't think it's the best solution ... but still better than the option 2

sapiens-sapide:
Option 1 enforced by PR #683

PR #683, opened 2017-11-29
Validate recovery_email provided at user creation.

sapiens-sapide:
+ must be a valid email address
+ must be unique
+ domain must be outside of instance's domains

Will prevent issue #673 to happen.

### DIFF
--- a/src/backend/interfaces/REST/py.server/caliopen_api/base/exception.py
+++ b/src/backend/interfaces/REST/py.server/caliopen_api/base/exception.py
@@ -34,22 +34,42 @@ class AuthenticationError(HTTPClientError):
 
     code: 401, title: Unauthorized
     """
+
+    def __init__(self, error=None):
+        if error:
+            self.message = error.message
+
     code = 401
     title = 'Authentication error'
     explanation = 'Wrong credentials (e.g., bad password or username)'
 
 
-class AuthorizationError(HTTPClientError):
-    code = 403
-    title = 'Forbidden'
-    explanation = 'Access was denied to this resource.'
+class AuthenticationError(HTTPClientError):
+    """
+    subclass of :class:`~HTTPClientError`
 
-    def __init__(self, detail=None, headers=None, comment=None,
-                 body_template=None, result=None, **kw):
-        HTTPClientError.__init__(self, detail=detail, headers=headers,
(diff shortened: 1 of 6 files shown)


src/backend/interfaces/REST/py.server/caliopen_api/base/exception.py
```diff
@@ -34,22 +34,42 @@ class AuthenticationError(HTTPClientError):
 
     code: 401, title: Unauthorized
     """
+
+    def __init__(self, error=None):
+        if error:
```
gdchamal: error may have no .message attribute and it's absolutely not a normal way to build an exception (give another exception on instanciation)

sapiens-sapide: I copied code from pyramid's httpexceptions package to build bespoke exception as subclass of HTTPClientError class.
Regarding the `message` property, it is a default property of python BaseException object, and we use this property to build the json error payload returned to client.

gdchamal: If it's a subclass of HTTPClientError (HTTPException as parent) you must change the __init__ constructor only for correct reasons, hence you break the hierarchy and polymorphism of your classes, at least.

And error in your __init__ can be many things other than an exception, so error.message may not exist.

Something is not correct in the object design

sapiens-sapide: so, if I want to be able to handle an `error` passed as an argument when calling the HTTPException, how do I must code it ? Something like this ? : 

```
    def __init__(self, error=None, detail=None, headers=None, comment=None,
                 body_template=None, **kw):
        if error is not None and not insinstance(error, Exception):
            raise ValueError("HTTP Unprocessable need an error ")
        super(_HTTPClientError, self).__init__(
            detail=detail, headers=headers, comment=comment,
            body_template=body_template, error=error, **kw)
```

gdchamal: Why do you want to handle error in the exception constructor ? Just handle it on callee when instanciate the exception.

sapiens-sapide:
commit #21d6482 change the way exceptions are called, with details about the error

PR #693, opened 2017-12-07
[wip]Feature/backend/python3

gdchamal:
Update caliopen python stack to use only pyhon version >= 3.4

It will permit to use asynchronous processing (send NATS messages) from API easily


gdchamal:
I need to prepare update of alpha platform before merging this branch.

Current testing on docker stack is OK

gdchamal:
Too much conflict and we're not ready enough for python3

Issue #611, opened 2017-11-03
Login doesn't work if i registred the credentials in my browser


AlxxxlA:
Hi,

I've registred the credentials for https://alpha.caliopen.org/ on my browser. If I disconnect then try to reconnect with the login and password autofilled, an error is diplayed 

![selection_046](https://user-images.githubusercontent.com/5237496/32370186-5dfd58ca-c08c-11e7-8e05-e90c61f7710d.png)

To reproduce : 
- Login on Caliopen and save the credentials in the browser
- Disconnect from Caliopen
- Try to reconnect with the credentials autofill by the browser

I'm using Firefox ESR 52.4.0

AlxxxlA:
Already reported on feedback : https://feedback.caliopen.org/t/reconnexion-impossible/123/2

Relative to : https://github.com/facebook/react/issues/1159

iamdey:
it's ok to keep it open until we have a fix

myleneb:
@iamdey  I tried to fix it here: https://github.com/CaliOpen/Caliopen/pull/697
based on this tip: https://medium.com/@dschmidt1992/auto-fill-with-redux-forms-9b51ad8ef962

myleneb:
Fixed with https://github.com/CaliOpen/Caliopen/pull/697

PR #697, opened 2017-12-08
Get inputs' values when browser autofills fields

myleneb:
this checks if inputs' values are changing after component did mount (ex. if browser auto-fills fields).
- also refactor InputText in order to pass `ref` prop.

Fix https://github.com/CaliOpen/Caliopen/issues/611

### DIFF
--- a/src/frontend/web_application/src/components/form/InputText/index.jsx
+++ b/src/frontend/web_application/src/components/form/InputText/index.jsx
@@ -19,7 +19,7 @@ class InputText extends PureComponent {
     bottomSpace: false,
     hasError: false,
     className: null,
-    inputRef: null,
+    inputRef: undefined,
   };
 
   state = {}
--- a/src/frontend/web_application/src/components/form/InputText/index.jsx
+++ b/src/frontend/web_application/src/components/form/InputText/index.jsx
@@ -1,35 +1,52 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import './style.scss';
 
-const InputText = ({ expanded, theme, bottomSpace, className, hasError, ...props }) => {
-  const inputTextClassName = classnames(
-    'm-input-text',
-    {
-      'm-input-text--expanded': expanded,
-      'm-input-text--light': theme === 'light',
-      'm-input-text--bottom-space': bottomSpace,
-      'm-input-text--error': hasError,
-    },
-    className
-  );
+class InputText extends PureComponent {
+  static propTypes = {
+    expanded: PropTypes.bool,
+    theme: PropTypes.string,
+    bottomSpace: PropTypes.bool,
+    hasError: PropTypes.bool,
+    className: PropTypes.string,
+    inputRef: PropTypes.func,
+  };
 
-  return (
-    <input
-      type="text"
-      className={inputTextClassName}
-      {...props}
-    />
-  );
-};
+  static defaultProps = {
+    expanded: false,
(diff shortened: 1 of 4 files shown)


src/frontend/web_application/src/components/form/InputText/index.jsx
```diff
@@ -1,35 +1,52 @@
...
-    className
-  );
+class InputText extends PureComponent {
+  static propTypes = {
+    expanded: PropTypes.bool,
+    theme: PropTypes.string,
+    bottomSpace: PropTypes.bool,
+    hasError: PropTypes.bool,
+    className: PropTypes.string,
+    inputRef: PropTypes.func,
```
iamdey: I'm not sure about this, by default input ref property gets a null instead of a function. It has more sense to set undefined as default value (like when it is not specified if I'm correct): `<input ref={undefined} /> === <input />`.

And well actually we do the same using `props` that have been transferred to input:

```jsx
<InputText name="foo" ref={inputRef => this.inputFoo = inputRef} />
```

 but it is not well named. I saw a more declarative syntax that should be better:

```jsx
<InputText inputProps={{ name: 'foo', ref: inputRef => this.inputFoo = inputRef }} />
```

What do you think ?

(in order to not have breaking changes or massive refactor, we should keep the actual props transmission: `<input {...inputProps} {...props} />`)

src/frontend/web_application/src/scenes/Signin/components/SigninForm/presenter.jsx
```diff
@@ -37,10 +37,25 @@ class SigninForm extends Component {
     this.setState(generateStateFromProps(this.props));
   }
 
+  componentDidMount() {
+    setTimeout(this.setValues(), 1);
```
iamdey: is timeout useful regarding to the blog post you pointed to ?

src/frontend/web_application/src/scenes/Signin/components/SigninForm/presenter.jsx
```diff
@@ -37,10 +37,25 @@ class SigninForm extends Component {
...
   componentWillReceiveProps(newProps) {
     this.setState(generateStateFromProps(newProps));
   }
 
+  setValues = () => {
+    const password = this.passwordInputRef.value;
+    const username = this.usernameInputRef.value;
+    this.setState(prevState => ({
+      formValues: {
+        password: prevState.password !== password && password,
```
iamdey: in case prevState.password & password are equal, this will evaluate `setState{{formValues: { password: false ... } })` which is incorrect.

this should be good:

```
this.setState(prevState => ({
  formValues: {
    password,
...
}));
```

src/frontend/web_application/src/scenes/Signin/components/SigninForm/presenter.jsx
myleneb: that's because setting state in `componentDidMount` is not possible without `setTimeout()`

src/frontend/web_application/src/components/form/InputText/index.jsx
myleneb: ok about `inputRef` default to `undefined`.
I don't get what you mean about the `ref` prop, is it only related to syntax? adding a `inputRef` prop is the right way to pass the `ref` prop to child component according to docs.

src/frontend/web_application/src/scenes/Signin/components/SigninForm/presenter.jsx
myleneb: yes of course /o\

PR #710, opened 2017-12-12
Feature/backend/is received

gdchamal:
Add a is_received boolean to message structure to permit filtering in an out messages easily.


gdchamal:
Need a script to populate alpha messages when structure will be altered

PR #717, opened 2017-12-13
refactor route config and global layout

iamdey:
1st step for refactor the page

I'm still not convinced by the interest of duplication of Page layout in each scenes

Take a look to Auth section, a simple decorator in route config also works great


iamdey:
https://stackoverflow.com/questions/47811835/how-to-organize-react-connected-components/47813075#47813075

PR #718, opened 2017-12-13
Add a CALIOPEN tag on inbound message that are from an internal domain sender

gdchamal:
A first tag CALIOPEN will appear automatically on incoming message from a sender defined in configuration file (hosted domains).

This should help validate UX/UI around tag display

### DIFF
--- a/src/backend/components/py.pi/caliopen_pi/qualifiers/mail.py
+++ b/src/backend/components/py.pi/caliopen_pi/qualifiers/mail.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 """Caliopen user message qualification logic."""
 from __future__ import absolute_import, print_function, unicode_literals
+
 import logging
+from datetime import datetime
 from caliopen_main.message.parameters import (NewInboundMessage,
                                               Participant,
                                               Attachment)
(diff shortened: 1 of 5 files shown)


LaurentChemla:
\o/

src/backend/components/py.pi/caliopen_pi/qualifiers/mail.py
```diff
@@ -44,7 +48,20 @@ def __init__(self, user):
...
         """Evaluate user rules to get all tags for a mail."""
-        return []
+        tags = []
+        if message.privacy_features.get('is_internal', False):
+            # XXX do not hardcode the wanted tag
+            internal_tag = [x for x in self.user.tags if x.name == 'CALIOPEN']
+            if internal_tag:
+                # XXX need to transform UserTag to ResourceTag
+                tag = ResourceTag()
+                tag.date_insert = datetime.utcnow()
```
sapiens-sapide: I'm not on the same wavelength… `date_insert` is the date the tag was created/inserted in db, not the date it was nested into the object.
Furthermore, we'll change the model: `tags` in messages, contacts, etc. will be an array of IDs, not an object.

gdchamal: Ok I'll wait for your update on nested tag structure to go further.

sapiens-sapide:
Update : new tag API must have a **lowercase** tag.name, not `CALIOPEN`

gdchamal:
[Auto tag caliopen internal messages](https://trello.com/c/1kYMvVSM/7-auto-tag-caliopen-internal-messages)

sapiens-sapide:
« caliopen » is not a generic name for this kind of system label. Should be « internal » or something else that can be use by other instances than Caliopen ones.  
There is the Tag's `label` property to put « caliopen ».

PR #747, opened 2018-01-04
[wip]Feature/backend/index sharding part1

gdchamal:
Currently an index is created per user, it's not a scalable design for elasticsearch and we must distribute differently data in index store.

We use a property User.shard_id to allocate many user per index and filter strictly per user_id when querying these indices.


gdchamal:
Will proceed differently than this solution, too heavy

Issue #630, opened 2017-11-09
import vcard ignore fn property


gdchamal:
Problem described here:
https://feedback.caliopen.org/t/champs-manquants-lors-de-limport-de-contacts/178

PR #751, opened 2018-01-11
Feature/backend/vcard v2

gdchamal:
Refonte du parsing vcard pour preparer a une meilleure algorithmie a l'import (deduplication, merge)

Sould fix: https://github.com/CaliOpen/Caliopen/issues/630


gdchamal:
[refonte parsing vcard](https://trello.com/c/h8PgSWZt/15-refonte-parsing-vcard)

src/backend/main/py.main/caliopen_main/contact/parsers/vcard.py
```diff
@@ -0,0 +1,160 @@
...
+        if 'TYPE' in param.params and param.params['TYPE']:
+            phone_type = param.params['TYPE'][0].lower()
+            if phone_type in _vcard_types:
+                phone.type = _vcard_types[phone_type]
+            else:
+                phone.type = 'other'
+        if 'PREF' in param.params:
+            phone.is_primary = True
+        try:
+            number = phonenumbers.parse(phone.number, None)
```
sapiens-sapide: by adding "fr" instead of None, the parser will fallback to a french phone number if — and only if — it did'nt find any country prefix.  
I think it is a good ideaa to add "fr" for now because we have mainly french users and it drastically raises the number of phones correctly normalized.  
We could add a `TODO` to remember to change this "fr" to a function that will try to find the most relevant country that fit for current user. (I did that on GO side)

gdchamal: I'll do a second parsing phase after having postal addresses to not have a 'fr' default value.
Idea is to normalize all phone numbers that are not exploiting other informations (if a postal address with a country it's great)

A default 'fr' value or any other default value is more dangerous than valuable as we will get lot of normalized values that may be are invalid

sapiens-sapide: as far as I understand the parser will not try to build a french phone number — even with "fr" default — if it leads to an invalid phone number.

gdchamal: Yes we normalize what can be normalized not guessed with a default value.
Country related stuff is client side when user fill a phone number. Value will be +(country)number formatted and will be normalized without any country value.

Issue #745, opened 2018-01-03
POST a tag with a label with upercases is always lowercased


iamdey:
POST /api/v2/tags

```
{"name":"loiso","label":"Loiso"}
```

GET  /api/v2/tags
```
{
  total: 1,
  tags: [
    {"name":"loiso"
    ,"date_insert":"2018-01-03T14:46:32.656Z","importance_level":0
    ,"label":"loiso"
    ,"type":"user"
  }]
}
```

sapiens-sapide:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix tag label issue](https://trello.com/c/uMCWf8vg/12-fix-tag-label-issue)

iamdey:
Also when trying to patch the label (same characters, different cases) it fails:

```
{"current_state":{"label":"yolo"},"label":"Yolo"}
```
```
{"errors":[{"code":422,"message":"[RESTfacility] tag's name/label conflict with existing one","name":""}]}
```

sapiens-sapide:
@iamdey : I'll fix the issue regarding the casse.  
Whereas, your second statement is not an issue. This is the way the API works for now: the `name` of the tag is the primary key used to embed tag's reference into objects. Consequently it must not change.

iamdey:
Does this will change ?

Because what happens if I try to create a tag with a label which already exists when "sluggyfied" as a name ?

(do you remember what we have statued last time ?)

sapiens-sapide:
I agree with you, this'll lead to a bad user experience. We need to discuss the point with @gdchamal because, for now, we decided to not take this usecase into account.

Issue #749, opened 2018-01-08
err 500 instead of 404 on patch message tags


iamdey:
Err. when trying to patch tags on a message that doesn't exist:

PATCH /api/v2/messages/missing/tags

```
{"tags":["nouveau1"],"current_state":{}}
```

response:

```
{"errors":[{"code":500,"message":"not found","name":""}]}
```

sapiens-sapide:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix tag label issue](https://trello.com/c/uMCWf8vg/12-fix-tag-label-issue)

PR #753, opened 2018-01-11
Fix issue #745 and issue #749 regarding tags name/label conflicts.

sapiens-sapide:


sapiens-sapide:
[Fix tags API issues](https://trello.com/c/uMCWf8vg/12-fix-tags-api-issues)

PR #764, opened 2018-01-16
Feat/frontend/bright message selector

myleneb:
In Timeline scene:
-Add/remove messages from selected list (MessageSelector);
-use AuthorAvatar as label for checkbox:

### DIFF
--- a/src/frontend/web_application/src/scenes/Timeline/components/MessageItem/presenter.jsx
+++ b/src/frontend/web_application/src/scenes/Timeline/components/MessageItem/presenter.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { v1 as uuidV1 } from 'uuid';
 import Moment from 'react-moment';
 import { Trans } from 'lingui-react';
 import Link from '../../../../components/Link';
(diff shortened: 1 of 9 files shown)


src/frontend/web_application/src/scenes/Timeline/components/MessageItem/presenter.jsx
```diff
@@ -151,7 +168,7 @@ class MessageItem extends Component {
             {this.renderDate()}
           </div>
           <div className="s-message-item__col-select">
-            <input type="checkbox" onChange={this.onCheckboxChange} />
+            <input type="checkbox" onChange={this.onCheckboxChange} id={id} />
```
iamdey: its value is missing for example I select from avatar in small view then I resize the window, the selection should be kept.

myleneb: Can you check last commit https://github.com/CaliOpen/Caliopen/pull/764/commits/8e98188b71510187185d098c5ab7decf118712d2 
I fixed this by passing `isMessageSelected` prop from parent `Timeline`  instead of managing it via `isSelected` state in `MessageItem`

iamdey: 👍

PR #767, opened 2018-01-17
Fix Dropdown to close on click

myleneb:


src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -87,14 +87,11 @@ class Dropdown extends Component {
     this.toggle(this.props.show);
 
     this.unsubscribeClickEvent = addEventListener('click', (ev) => {
-      if (!this.props.closeOnClick) {
```
iamdey: I've specifically add this fix here: https://github.com/CaliOpen/Caliopen/commit/7483a30afb3c2806dd7d41057ef40be46a361680

because when closeOnClick is false (default) and we click somewhere on the page except the dropdown control, `dropdownClick` is set to false (and so `controlClick` and `exeptRefsClick`) ; so the dropdown is closed.

May be it is better to not register at all the click event when `closeOnClick` prop is false.

iamdey: You can actually see this behavior when selecting a tag on a message or a contact.

myleneb: If we add a `closeOnClickExeptSelf` prop this seems to work well for all cases:
```
    this.unsubscribeClickEvent = addEventListener('click', (ev) => {
      const target = ev.target;

      const dropdownClick = this.props.closeOnClickExeptSelf &&
        (this.dropdown === target || this.dropdown.contains(target));
      const controlClick = this.dropdownControl &&
        (this.dropdownControl === target || this.dropdownControl.contains(target));

      if (controlClick) {
        this.toggle(!this.state.isOpen);

        return;
      }
      if (!this.props.closeOnClick) { return; }
      if (dropdownClick) { return; }


      this.toggle(false);
    });
```

btw this example removes `closeOnClickExeptRefs` prop.
Passing refs to siblings seems not to be a good way to handle exceptions.
Maybe should `Dropdown` only manages self click events, and click exceptions should be handle in parent component (like in `RecipientList`: exception for `this.searchInputRef` is handle in `handleSearchInputFocus()` and exception for ref `this.recipientListRef` which should be pass to `Dropdown` via `closeOnClickExceptRefs` doesn't seem to work.)

or am I very wrong?

iamdey: About recipientList, you are right about searchInput but I'm not sure about the recipientList.
Eventually without `closeOnClickExeptRefs` it can works with some tunning, so I'm okay with this.

I'm afraid but I'm not sure that `closeOnClick` and `closeOnClickExeptSelf` are incompatible in some cases.

May be a single property for both behavior should be better; what do you think ?

And when `closeOnClick` is false do not register click event listener at all.

myleneb: I tested it on ImportanceLevel slider and it works fine with both true `closeOnClick` and `closeOnClickExeptSelf` props, but you're right it will be smarter with one only prop.
something like:
```
static propTypes = {
  closeOnClick: PropTypes.oneOf(['none', 'all', 'exeptSelf']),
}
static defaultProps = {
  closeOnClick: 'none',
}
```
?

about `closeOnClickExeptRefs` it seems it doesn't always work because of react mounting cycle (I think).

myleneb: mm maybe rather:
```
static propTypes = {
  closeOnClick: PropTypes.oneOf(['all','exeptSelf']),
}
static defaultProps = {
  closeOnClick: null,
}
```

iamdey: 👍

iamdey: `exceptSelf`

src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -48,30 +48,31 @@ class Dropdown extends Component {
...
     // <div ref={foo => this.foo =foo} />
     // <Dropdown closeOnClickExceptRefs={[this.foo]} />
-    closeOnClickExceptRefs: PropTypes.arrayOf(PropTypes.shape({})),
+    // closeOnClickExceptRefs: PropTypes.arrayOf(PropTypes.shape({})),
     closeOnScroll: PropTypes.bool, // should Dropdown close on windows scroll ?
     isMenu: PropTypes.bool,
     position: PropTypes.oneOf(['top', 'bottom']),
     onToggle: PropTypes.func,
     show: PropTypes.bool,
+    exeptionClickFunc: PropTypes.func,
```
iamdey: `exceptionClickFunc`

src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -48,30 +48,31 @@ class Dropdown extends Component {
...
     children: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
     className: PropTypes.string,
-    closeOnClick: PropTypes.bool, // should Dropdown close on click?
+    closeOnClick: PropTypes.oneOf(['all', 'exceptSelf']), // should Dropdown close on click?
     // closeOnClickExceptRefs: array of refs that should not close dropdown on click
     // usage:
     // <div ref={foo => this.foo =foo} />
     // <Dropdown closeOnClickExceptRefs={[this.foo]} />
-    closeOnClickExceptRefs: PropTypes.arrayOf(PropTypes.shape({})),
+    // closeOnClickExceptRefs: PropTypes.arrayOf(PropTypes.shape({})),
```
iamdey: you can remove old commented code

src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -48,30 +48,31 @@ class Dropdown extends Component {
     alignRight: PropTypes.bool, // force align right
     children: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
     className: PropTypes.string,
-    closeOnClick: PropTypes.bool, // should Dropdown close on click?
+    closeOnClick: PropTypes.oneOf(['all', 'exceptSelf']), // should Dropdown close on click?
```
iamdey: I like to add constants to make sure the string has no typo error:

```
const CLOSE_ON_CLICK_ALL = 'all';
const CLOSE_ON_CLICK_EXCEPT_SELF = 'exceptSelf';
```

src/frontend/web_application/src/components/Dropdown/index.jsx
```diff
@@ -86,26 +87,31 @@ class Dropdown extends Component {
...
+      this.unsubscribeClickEvent = addEventListener('click', (ev) => {
+        const target = ev.target;
+        // const exceptRefs = this.props.closeOnClickExceptRefs;
 
-      const dropdownClick = !this.props.closeOnClick &&
-        (this.dropdown === target || this.dropdown.contains(target));
-      const exeptRefsClick = exceptRefs && exceptRefs.find(ref => (ref === target));
-      const controlClick = this.dropdownControl &&
-        (this.dropdownControl === target || this.dropdownControl.contains(target));
+        const dropdownClick = (this.props.closeOnClick === 'exceptSelf') &&
```
iamdey: I prefer when the variable describe correctly what happens: if the click is on the dropdown but closeOnClick is something else, `dropdownClick` is set to false that is incorrect.

I think it's better to add this later in the condition (or rename the variable to something very weird):

```
if (this.props.closeOnClick === 'exceptSelf' && dropdownClick) { return; }
```

PR #778, opened 2018-01-26
Refactoring/backend/contact apiv1 to v2

sapiens-sapide:
Many improvements in this APIv2, notably : 
* control rights on each properties in patch to prevent modification of protected properties
* update lookup and related tables in stores.

APIv2 should work out-of-the-box with frontend, just need to change url.

Slight modifications have been made on index mapping for Contact and on Contact's model (field deleted changed from a boolean to a date) => migrations scripts have to been written.


sapiens-sapide:
[Contacts API v1-> v2](https://trello.com/c/1QyzvyAn/8-contacts-api-v1-v2)

iamdey:
it sounds like apiv2 for contacts works great but I got this on messages tags:

```
PATCH /api/v2/messages/b0d83aa2-bbb0-4d60-a41e-4b187d10dca9/tags
payload:
{"tags":["inbox"],"current_state":{}}

response
{"errors":[{"code":0,"message":"[RESTfacility] UpdateResourceTags : helpers.UpdateWithPatch failed","name":""},{"code":5,"message":"current state not consistent with db state","name":""}]}
```

no problems with contact tags

iamdey:
a little thing that makes a big difference, before the response of POST contact gives `{ location: '....`

and now it is `{ Location: '....`

_At the end I would prefer to receive an id instead of a location (or may be both for strange cases)_

gdchamal:
go merge, c'est trop lourd pour une review correcte

PR #779, opened 2018-01-26
Feat/frontend/bright forms

myleneb:
This put bright styles on forms and related components.
- [x] new Checkbox
- [x] Settings forms
- [x] TagsSettings forms
- [x] Contact forms
- [ ] Signin/Signup forms
- [ ] User settings forms

Also, "brighted" form components are put in `/brightForm` and will be moved to `/form` when migration dark => bright will be OK.


src/frontend/web_application/src/components/brightForm/Checkbox/index.jsx
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import './style.scss';
+
+class Checkbox extends Component {
```
iamdey: about structure: https://pad.caliopen.org/p/wd0dlobnQ9

src/frontend/web_application/src/components/brightForm/Checkbox/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+  static defaultProps = {
+    indeterminate: null,
+    showLabelforSr: false,
+  };
+
+  state = {}
+
+  componentDidMount() {
+    // this sets indeterminate attribute:
```
iamdey: may be something more specific, something like: 

> apply the indeterminate attribute of the real checkbox element

src/frontend/web_application/src/scenes/ApplicationSettings/components/ContactSettings/presenter.jsx
```diff
@@ -44,22 +44,22 @@ class ContactsForm extends Component {
     return (
       <FormGrid className="m-contacts-form">
         <FormRow>
-          <FormColumn size="shrink" bottomSpace >
+          <FormColumn rightSpace={false} bottomSpace >
```
iamdey: I don't like much the `rightSpace` modifier

no idea how to do it the "right way" for now

src/frontend/web_application/src/components/brightForm/CheckboxFieldGroup/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { v1 as uuidV1 } from 'uuid';
+import Checkbox from '../Checkbox';
+import Switch from '../../form/CheckboxFieldGroup/components/Switch';
+import { FieldErrors } from '../';
+
+import './style.scss';
+
+class CheckboxFieldGroup extends PureComponent {
```
iamdey: What do you think about changing the way we build input form, may be something like:

```
<FieldGroup errors={errors}><Checkbox ... /><Label ... /></FieldGroup>
```

src/frontend/web_application/src/components/Button/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import Icon from '../Icon';
+import { Icon } from '../';
```
iamdey: this is a thing we didn't talk.

I found that sometimes the import doesn't work using a directory index.

for example if three components are in the same directory like Menu, Button & Icon, the import will work for Menu using Icon because Icon has already  been parsed by webpack and it won't for Button using Icon.

It may change with versions of webpack, babel etc. 

Well I think to make sure this works is to not using import from index when we are in the directory.

**OK:**

```
// file: src/scene/Foo/presenter.jsx
import { Button } from '../components';

// file: src/components/Button/index.jsx
import Icon from '../Icon';
```

**NOT OK**
```
// file: src/scene/Foo/presenter.jsx
// It works correctly but not efficient
import Button from '../components/Button';

// file: src/components/Button/index.jsx
// It might be completely broken
import { Icon } from '../';
```

myleneb: ok

myleneb: and what about sub-directories of `/components/`?
Ex. for file: `src/components/Foo/components/Bla/presenter.jsx`

PR #787, opened 2018-02-05
Refacto/frontend/directories

myleneb:
This tries to move some components into their own scene/layout/module.
Missing: components related to ~~draft~~, ~~tagsForm~~ and avatar.


myleneb:
I moved `MultidimensionalPi` into `modules/pi` to avoid error on mixed imports in `MessageList`.
See commit https://github.com/CaliOpen/Caliopen/pull/787/commits/c8f85ff0325751e8921f7cdce4cced8405cd768e

iamdey:
still ok but I've changed some things in my previous emrged PR

PR #798, opened 2018-02-12
Fix Modal to prevent warning about appElement

myleneb:
Also fix body styles when modal is open to avoid scrolling.
ref: https://github.com/reactjs/react-modal#app-element

### DIFF
--- a/src/frontend/web_application/src/components/Modal/index.jsx
+++ b/src/frontend/web_application/src/components/Modal/index.jsx
@@ -9,6 +9,7 @@ const Modal = ({ className, title, children, onClose, ...props }) => (
   <ReactModal
     className={classnames('m-modal', className)}
     overlayClassName="m-modal__overlay"
+    appElement={document.body}
     {...props}
   >
     <header className="m-modal__header">
--- a/src/frontend/web_application/src/components/Modal/style.scss
+++ b/src/frontend/web_application/src/components/Modal/style.scss
@@ -3,6 +3,14 @@
 
 $m-modal__overlay-background: rgba(0, 0, 0, .75);
 
+body {
+  &.ReactModal__Body--open {
+    // this is the class applied by ReactModal to body when modal is open
+    // add overflow: hidden to avoid scrolling on body then
+    overflow: hidden;
+  }
+}
+
 .m-modal {
   position: absolute;
   top: 50%;


src/frontend/web_application/src/components/Modal/index.jsx
```diff
@@ -9,6 +9,7 @@ const Modal = ({ className, title, children, onClose, ...props }) => (
   <ReactModal
     className={classnames('m-modal', className)}
     overlayClassName="m-modal__overlay"
+    appElement={document.body}
```
iamdey: I think using `document` can be a problem when rendering server side, using a css selector should be better. 
And  add a comment to explain why it is important (useful to limit the screenreader to the content of modal)

src/frontend/web_application/src/components/Modal/style.scss
```diff
@@ -3,6 +3,14 @@
 
 $m-modal__overlay-background: rgba(0, 0, 0, .75);
 
+body {
+  &.ReactModal__Body--open {
+    // this is the class applied by ReactModal to body when modal is open
+    // add overflow: hidden to avoid scrolling on body then
```
iamdey: what if the modal is taller as screen height ?

myleneb: imo Modal should not be tallest than screen height and manage their own overflow, no?

src/frontend/web_application/src/components/Modal/index.jsx
myleneb: ok, but which selector then?

iamdey: `"#root"` should be the selector to use I suppose.

https://github.com/CaliOpen/Caliopen/blob/master/src/frontend/web_application/src/index.jsx#L43

src/frontend/web_application/src/components/Modal/index.jsx
```diff
@@ -1,40 +1,48 @@
...
 
-Modal.defaultProps = {
-  className: undefined,
-  title: undefined,
-  children: undefined,
-  onClose: () => {},
-};
+  render() {
+    const { className, title, children, onClose, ...props } = this.props;
+    const rootElement = document.getElementById('root');
```
iamdey: express server will throw an error here (it executes componentWillMount and render)

src/frontend/web_application/src/components/Modal/index.jsx
```diff
@@ -1,40 +1,48 @@
...
+
+    return (
+      <ReactModal
+        className={classnames('m-modal', className)}
+        overlayClassName="m-modal__overlay"
+        appElement={rootElement}
```
iamdey: i think a simple `appElement="#root"` will work

to be confirmed

src/frontend/web_application/src/components/Modal/index.jsx
iamdey: (it doesn't because we didn't make it work correctly for some reasons : translation provider and may be some other components)

src/frontend/web_application/src/components/Modal/index.jsx
myleneb: no this doesn't work :( `appElement` prop must be of type "instance of HTMLElement", not string.

iamdey: I don't find when it has been changed but looks better now : http://reactcommunity.org/react-modal/examples/set_app_element.html#using-setappelement

PR #800, opened 2018-02-14
Feature/backend/crud devices api

sapiens-sapide:


src/backend/defs/go-objects/device.go
```diff
@@ -0,0 +1,275 @@
...
+		close(getSet)
+		d.Locker.Unlock()
+	}(d.Locker, getSet)
+
+	return getSet
+}
+
+// IsValidDeviceType checks the `t` string against the constant `DeviceTypes`
+// returns true if `t` is a valid type for a device.
+func IsValidDeviceType(t string) bool {
```
gdchamal: Is there not a way to define like an enumeration, instead of a delimited string ?

sapiens-sapide: no `enum` type in GO. See https://github.com/golang/go/issues/19814 and the response from lance taylor : « A new language feature needs compelling use cases. All language features are useful, or nobody would propose them; the question is: are they useful enough to justify complicating the language and requiring everyone to learn the new concepts? What are the compelling use cases here? How will people use these? For example, would people expect to be able to iterate over the set of valid enum values, and if so how would they do that? Does this proposal do more than let you avoid adding default cases to some switches? »

gdchamal: So forget enum name, think a map or a structure of valid values. Not a single string like this.

sapiens-sapide: Could be a `map[string]bool`, but map are unordered ; Thus there is no way with a map to find the `default` string for device.  
Also, making it a simple string makes it easy to put it in a config file, easying edit of devices' types list.

sapiens-sapide: Changed to an array in commit below

gdchamal:
LGTM

PR #804, opened 2018-02-15
Fix/frontend/timeline multiple delete

myleneb:


src/frontend/web_application/src/scenes/Timeline/components/MessageSelector/index.jsx
```diff
@@ -74,9 +75,10 @@ class MessageSelector extends Component {
           <Checkbox
             label={i18n._('message-list.action.select_all_messages', { defaults: 'Select/deselect all messages' })}
             id="message-selector"
-            defaultChecked={checked}
+            checked={checked}
```
myleneb: changed `defaultChecked` to `checked` prop because `defaultChecked` doesn't update when all displayed messages are selected and then deleted. Works fine with `checked` prop.

PR #805, opened 2018-02-15
Add RFC for API notifications and swagger definition for matching route.

sapiens-sapide:


sapiens-sapide:
[Notification nouveau message](https://trello.com/c/d8IGYj8z/32-notification-nouveau-message)

doc/RFCs/APInotification.md
```diff
@@ -0,0 +1,245 @@
...
+      warning: sending email failed  
+      code: 23  
+      message: MTA timeout  
+```
+
+##### notification que l'index est à jour (suite à create/update de document)
+
+l'utilisateur a modifié un contact (`PATCH /contacts/{contact_id}`)
+
+lors de son prochain call sur GET /v2/notifications, le frontend pourrait recevoir la notification suivante :
```
iamdey: -> utile pour invalider les données cachées dans le store redux 
    je verrais plutôt `to : entity-store, type: event`

    "event-manager" c'est trop abstrait

doc/RFCs/APInotification.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+
+#### Exemples de notifications :
+
+##### notification de réception de email(s) :
+
+```yaml  
+notifications:  
+  - from: email-facility
```
gdchamal: everything is a facility, so I don't think it's needed to have this word in description

doc/RFCs/APInotification.md
```diff
@@ -0,0 +1,245 @@
...
+### Modèle de données :
+
+La table des notifications est stockées dans cassandra.
+
+#### Chaque `INSERT` se fera avec la directive `USING TTL xx` pour gérer le nettoyage automatique de la table
+
+```sql  
+CREATE TABLE user_notification (  
+user_id uuid,  
+timestamp timestamp,
```
gdchamal: device_id is may be needed to identity for which device the notification is

iamdey:
As described here https://trello.com/c/d8IGYj8z/32-notification-nouveau-message

We can see in the mockup that asynchronous actions requires a state that is not well described in this RFC.

For example for the contact import we can easily loose the state of the import. may be it can a fetch on a certain "notification" (so that's not a notification anymore) to get the state or the percentile progression of the action.

doc/RFCs/APInotification.md
sapiens-sapide: But, a notification is not always for a specific device.  
For example:  if a new email arrived while an user is disconnected, a notification is queued. But, we don't know yet which device the user will use to reconnect. So the notification is « for first device ». Isn't it ?

gdchamal: Yes you're right, notification will be seen by all actives devices by default.

gdchamal: Yes you're right. May they will have notification dedicated to specific devices, but later, I ensure at this time about that need.

sapiens-sapide:
branch has been merged into PR #819

PR #808, opened 2018-02-16
Refacto/frontend/avatar in modules

myleneb:
+ This moves all Avatar related components into `modules/avatar`.
+ also creates `AvatarLetterWrapper` to manage props ( size, isSelected...) of `ContactAvatarLetter` and `AuthorAvatarLetter`


src/frontend/web_application/src/modules/avatar/components/AvatarLetterWrapper/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+      [SIZE_LARGE]: 'm-avatar-letter-wrapper--large',
+      [SIZE_XLARGE]: 'm-avatar-letter-wrapper--xlarge',
+      [SIZE_XXLARGE]: 'm-avatar-letter-wrapper--xxlarge',
+    };
+    const classNameSize = classNameModifiers[size];
+    const letterClassNameSize = classNameSize ? `${classNameSize}__letter` : null;
+    const avatarLetter = React.Children.map(children, (child) => {
+      const avatarLetterClassName = child.props.className || '';
+
+      return React.cloneElement(child, {
```
iamdey: the jsx notation is mainly used https://reactjs.org/docs/react-api.html#cloneelement:

```
return (<child.type className={...} {...props} />);
```

PR #810, opened 2018-02-20
Feature/backend/pi message from contact

gdchamal:
Participants know as contacts in a message can increase PI.context of this message


gdchamal:
[PI message from PI contact](https://trello.com/c/7rQYA3RN/35-pi-message-from-pi-contact)

PR #814, opened 2018-02-22
Feature/backend/auth with new device

gdchamal:
When an user authenticate with a new device, we can/must provide new device information as it's done on signin


gdchamal:
[APIv1: Authentification avec contexte et device](https://trello.com/c/xNb8Thpb/21-apiv1-authentification-avec-contexte-et-device)

PR #815, opened 2018-02-26
handle multiple tags management on Timeline

iamdey:
+ refactor updateTagCollection using getState instead of passing *outdated* userTags
+ refactor WithSearchTags render prop component into a service


iamdey:
fast click management is better, not yet perfect

PR #816, opened 2018-02-28
Fixes/backend/devices api

sapiens-sapide:


src/backend/interfaces/REST/go.server/operations/devices/devices.go
```diff
@@ -8,24 +8,25 @@ package devices
...
-	"github.com/satori/go.uuid"
 	"io/ioutil"
 	"net/http"
 	"strconv"
-	"strings"
 )
 
 // NewDevice handles POST /devices
 func NewDevice(ctx *gin.Context) {
+	e := swgErr.New(http.StatusForbidden, "do not create new device this way")
```
gdchamal: remove the route from swagger, it will raise a 404, a forbidden is not a valid response

PR #817, opened 2018-02-28
Only the first device is named default

gdchamal:
As spooted by @sapiens-sapide , user can have many 'default' devices with previous logic.

Only the first one is named default, others will be named 'new device <device_type>'


sapiens-sapide:
👍

PR #826, opened 2018-03-14
Improvement/backend/attachments

sapiens-sapide:
improvements for : 
- https://trello.com/c/Imek7lAM
- https://trello.com/c/rJamYQVh

### DIFF
--- a/src/backend/defs/rest-api/objects/Attachment.yaml
+++ b/src/backend/defs/rest-api/objects/Attachment.yaml
@@ -1,13 +1,11 @@
 ---
 type: object
 properties:
-  attachment_id:
-    type: string
   content_type:
     type: string
   is_inline:
     type: boolean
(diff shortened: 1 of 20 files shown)


src/backend/defs/rest-api/objects/Attachment.yaml
```diff
@@ -1,13 +1,11 @@
 ---
 type: object
 properties:
-  attachment_id:
```
iamdey: hmm does it mean that we only use an index ?
What about this use case : 

1. clic DELETE attachments/0
2. clic DELETE attachments/1
3. DELETE 1 success
4. invalidate message
5. -> DELETE 2 will probably fail

sapiens-sapide:
[Add/delete attachment in a draft](https://trello.com/c/Imek7lAM/52-add-delete-attachment-in-a-draft)

PR #827, opened 2018-03-15
DevicesSettings styles

myleneb:
- remove devices NavList
- fix styles
- add locales


iamdey:
I've added wording about new device

capture d ecran - 2018-03-20 a 11 45 51

PR #832, opened 2018-03-26
Bright styles on Message in MessageList

myleneb:
This is a first pass for bright styles on MessageList

### DIFF
--- a/src/frontend/web_application/src/scenes/MessageList/components/Message/style.scss
+++ b/src/frontend/web_application/src/scenes/MessageList/components/Message/style.scss
@@ -58,10 +58,11 @@ $message_content_width: 45rem;
   }
 
   &__participant {
-    &::after { content: ', '}
+    display: inline-block;
+    padding-right: $co-component__spacing / 2;
 
     &:last-child {
-      &::after { content: ''}
+      padding-right: 0;
     }
   }
 
@@ -76,7 +77,6 @@ $message_content_width: 45rem;
   }
 
   &__subject {
-    display: block;
     margin: $co-margin 0;
     padding-bottom: $co-component__spacing;
     border-bottom: 1px solid $co-color__fg__border--higher;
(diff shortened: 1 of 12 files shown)


src/frontend/web_application/src/scenes/MessageList/components/Message/style.scss
```diff
@@ -1,160 +1,109 @@
...
-    border-left: 1px solid $co-color__fg__back--lower;
+    margin-left: auto;
   }
 
-  &__subject {
-    padding: $co-component__spacing;
-    border-bottom: 1px solid $co-color__fg__back--lower;
-    color: $co-color__fg__text--lower;
+  &__participant {
+    &::after { content: ', '}
```
iamdey: I'm not sure it a good idea,
this gives a responsibility of the content to the stylesheets which is initially the job of react and may be accessibility softwares might not interpret it (I have no idea).

iamdey: this what I meant : https://css-tricks.com/css-content/

1/ "content not design"
2/ "Browser support / Accessibility"

src/frontend/web_application/src/scenes/MessageList/components/Message/style.scss
```diff
@@ -1,160 +1,109 @@
...
-    border-top: 1px solid $co-color__fg__border--higher;
-  }
-
-  &__attachements-item {
-    @include flex-grid-row;
-    margin-bottom: $co-component__spacing;
-    background: $co-color__fg__border--higher;
-    font-size: $co-font__size--small;
+  &__subject {
+    display: block;
```
iamdey: why?

myleneb: `display: block` because of the 100% large border-bottom. Is it better to put a `div` instead of `span` for `m-message__subject` ? or maybe better a `h2` or any other title?

That make me think, we may better use an `<article>` tag to diplay each message
`<article id={message.message_id} className="m-message">`
and then `<header>` and `<footer>` tags.
What do you think?

iamdey: I think the more semantic the better
so :+1: 

http://devdocs.io/html/element/article

PR #836, opened 2018-03-28
fix delete draft button that was always disable

iamdey:


iamdey:
I suppose it fixes #843

PR #844, opened 2018-03-30
add the not found unicorn

iamdey:


iamdey:
I can't reproduce your issue @myleneb. 
what's your browser ?

myleneb:
Chromium Version 64.0.3282.186 and it also bugs with epiphany ( = WebKit)

iamdey:
I've checked those browsers. Is it still the case ?

iamdey:
I was able to see it on saucelabs

0021screenshot

myleneb:
There are browsers' related issues due to monospace font:

> Major web browsers automatically reduce the size of monospace text in a range of situations. They do this due to the greater width of many monospace typefaces in comparison to other fonts at the same text height. While well-intentioned, their handling of monospace fonts is fundamentally broken, and an unusual CSS rule is required to fix the problem.

http://code.iamkate.com/html-and-css/fixing-browsers-broken-monospace-font-handling/

the hack `font-family:monospace,monospace;` seems to work for my chromium issue but not for epiphany.

iamdey:
if we reduce the size of the font of this element, does it work fine ?

myleneb:
see https://github.com/CaliOpen/Caliopen/pull/876

PR #847, opened 2018-04-05
Fix contact.title default value in ContactBook

myleneb:
Fix error on `localeCompare` function when `title` is null or undefined.


iamdey:
you should be able to change base branch to master 
& add an entry in changelog

iamdey:
ok this is not good :/

iamdey:
re-PR in #849

PR #848, opened 2018-04-05
upgrade node version, update dev deps (webpack) & eslint

iamdey:
+ fix new eslint errors
+ restore stylelint


gdchamal:
Not yet, I expect bug fixes on alpha before upgrading node and check and debug a new release process

iamdey:
of course this is for the next big release

PR #884, opened 2018-04-25
refactor doc according to https://pad.caliopen.org/p/doc

iamdey:
https://pad.caliopen.org/p/doc


doc/install/semi-dockerized-installation.md
```diff
@@ -0,0 +1,142 @@
+# Semi dockerized installation
```
gdchamal: there is others kind of halk dockerized installation, not only frontent one.

iamdey: what do I choose ?
* "dockerized installation" (and I add an optional part to how to deconnect and run frontend natively ?) 
* "installation for frontend development"

Imo, this will be changed very soon thanks to the work of pablo

PR #895, opened 2018-05-03
Improvment/backend/draft date ordering

sapiens-sapide:
- Add the `date_sort` property to Message and logic to compute this date : 

> date_sort = date_insert for inbound messages
> date_sort = date for outbound messages

- Return messages list (`GET /messages`) ordered by `date_sort`
- Add 2 scripts to migrate index and cassandra store :  

```shell
cd devtools
python set_date_sort_in_db.py -f ../../src/backend/configs/caliopen.yaml.template
caliopen -f ../../src/backend/configs/caliopen.yaml.template migrate_index -s migrations/index_migration_v4_to_v5.py
```


devtools/migrations/index_migration_v4_to_v5.py
```diff
@@ -0,0 +1,198 @@
...
+            if self.es_client.indices.exists(old_index):
+                log.warn('Delete the empty index {} with old version found'.
+                         format(index))
+                self.es_client.indices.close(index)
+                self.es_client.indices.delete(index)
+        return True
+
+    def run(self):
+        indexes = self.es_client.indices.get("_all")
+        log.warn("found {} indexes".format(len(indexes)))
```
gdchamal: not a warning AFAIK

devtools/migrations/set_date_sort_in_db.py
```diff
@@ -0,0 +1,32 @@
...
+    args = parser.parse_args()
+    Configuration.load(args.conffile, 'global')
+    connect_storage()
+    from caliopen_main.message.store.message import Message
+
+    log.info("updating model for table message")
+    sync_table(Message)
+    log.info("setting date_sort for {} messages".format(Message.objects().count()))
+    for message in Message.all():
+        message.date_sort = message.date_insert if message.is_received \
```
gdchamal: If date_sort is already set (script previously run partially) don't update

iamdey:
does the discussion messages are also sorted the same way ?

sapiens-sapide:
if you get your discussion messages from route `/messages`, then yes.

PR #900, opened 2018-05-03
Feature/backend/better setup

gdchamal:
setup CLI command setup storage but also configure default TTL for notifications


sapiens-sapide:
LGTM

PR #905, opened 2018-05-03
Fix TagsForm width

myleneb:
This makes `TagsForm` width fixed to avoid `Modal` to resize while adding or removing tags.


src/frontend/web_application/src/modules/tags/components/TagsForm/style.scss
```diff
@@ -2,6 +2,8 @@
 @import '../../../../styles/vendor/bootstrap_foundation-sites';
 
 .m-tags-form {
+  width: 100%;
```
iamdey: witdh is required even if you set max-witdh ?

myleneb: yep that's useful for screens narrower than `max-width` (=380px)

PR #911, opened 2018-05-07
Fix/frontend/timeline align

myleneb:
- `MessageItem` alignments on `Timeline` (see https://trello.com/c/zXvubJCS/66-review-ux-timeline)
- `MessageSelector` only shows action buttons when one or more message is selected.

### DIFF
--- a/src/frontend/web_application/src/components/MessageDate/presenter.jsx
+++ b/src/frontend/web_application/src/components/MessageDate/presenter.jsx
@@ -18,8 +18,10 @@ class MessageDate extends PureComponent {
         title={dateTime.format('LLL')}
         dateTime={dateTime.format('')}
       >
-        { /* <span className="m-message-date__date">{dateTime.format('L')}</span> */}
-        { /* ' ' */}
+        {
+          // TODO: setup Moment to display date as defined by UI
+        /* <span className="m-message-date__date">{dateTime.format('L')}</span> */
+        }
         <span className="m-message-date__time">{dateTime.format('LT')}</span>
       </time>
     );
--- a/src/frontend/web_application/src/components/MessageDate/presenter.jsx
+++ b/src/frontend/web_application/src/components/MessageDate/presenter.jsx
@@ -7,8 +7,26 @@ class MessageDate extends PureComponent {
   static propTypes = {
     dateTime: PropTypes.instanceOf(Moment).isRequired,
   };
-  static defaultProps = {
-  };
+  static defaultProps = {};
+
+  renderDate = () => {
+    const { dateTime } = this.props;
+
+    const diffDays = Moment().diff(dateTime, 'days');
+    const diffYear = Moment().diff(dateTime, 'years');
+
+    switch (true) {
+      default:
+      case diffDays === 0:
+        return dateTime.format('LT'); // 11:00
+      case diffDays > 0 && diffDays <= 7:
+        return dateTime.format('dddd'); // monday
+      case diffYear === 0 && diffDays > 7:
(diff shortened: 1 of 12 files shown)


src/frontend/web_application/src/components/MessageDate/presenter.jsx
```diff
@@ -18,8 +18,10 @@ class MessageDate extends PureComponent {
         title={dateTime.format('LLL')}
         dateTime={dateTime.format('')}
       >
-        <span className="m-message-date__date">{dateTime.format('L')}</span>
-        {' '}
+        {
+          // TODO: setup Moment to display date as defined by UI
```
iamdey: yes ?

src/frontend/web_application/src/scenes/Timeline/components/MessageItem/style.scss
```diff
@@ -76,19 +76,23 @@ $s-message-item__pi-width: 7px;
...
-    font-size: $co-font__size;
     font-weight: 600;
   }
 
   &__topic {
     @include flex-grid-size(12);
+    font-weight: 400;
 
-    &--unread { font-weight: 600; }
+    // &--unread { font-weight: 600; }
```
iamdey: to remove ?

src/frontend/web_application/src/scenes/Timeline/components/MessageItem/style.scss
```diff
@@ -126,8 +126,12 @@ $s-message-item__pi-width: 7px;
 
     &__col-dates { @include flex-grid-size(shrink); }
 
-    &__author { @include flex-grid-size(shrink); }
+    &__author { @include flex-grid-size(3); }
+
+    // &__type-label { display: inline }
```
iamdey: to remove ?

src/frontend/web_application/src/scenes/Timeline/components/MessageItem/presenter.jsx
```diff
@@ -147,7 +153,7 @@ class MessageItem extends Component {
           }
         )}
       >
-        <div className="s-message-item__col-avatar">
+        <div className="s-message-item__column--avatar">
```
iamdey: hmm. this can be sort of modifier but I think not:

Ok all of these are columns but in this case the name of the modifier should be "shrink" or "large" or ...
But finally this doesn't give any advantages:

```scss
&__column { 
  // no generic rule
  &--avatar { // is the name of the column (not a modifier name)
    // specific rule 
  }
}
```

I prefer what was before and imo more in the concept of the Element in BEM:

```scss
&__col-<the name of the column> {
 // the directives of the column
}
```

Or... if you prefer to go to refactor (but overkill for now):

```
<Column className="s-message-item__avatar">....
```

---

Eventually if it is required for some reasons, please leave a comment in code explaining why we needed to break the "rules" (because that's ok ;-) )

src/frontend/web_application/src/scenes/Timeline/components/MessageItem/presenter.jsx
```diff
@@ -151,20 +162,13 @@ class MessageItem extends Component {
...
         </div>
-        <div className={classnames(
-          's-message-item__col-dates',
-          { 's-message-item__col-dates--unread': message.is_unread }
-          )}
-        >
+        <div className="s-message-item__column--date">
           {this.renderDate()}
         </div>
-        <div className="s-message-item__col-select">
```
iamdey: the tests fails because this className was used as a selector:

```
Failed: No element found using locator: By(css selector, .s-message-item__col-select input[type=checkbox])
```

PR #927, opened 2018-05-16
[WIP] HTTP request signature

MonsieurLanza:
Work in progress. As for now signs request from `axios-middleware` with url's path.


iamdey:
looks pretty good for now

MonsieurLanza:
Added serializer. Axios' one is encapsulated and unreachable without ugly tweaks. This one is mostly a refactor of the former, to accomodate our `.eslint`.

src/frontend/web_application/src/modules/device/services/signature/index.js
```diff
@@ -0,0 +1,27 @@
...
+import { getConfig } from '../storage';
+import { buildURL } from '../../../../services/url';
+
+const buildMessage = ({ url, params }) => {
+  const message = buildURL(url, params);
+
+  return hash.sha1(message);
+};
+
+export const signRequest = (messageParts, priv) => {
```
iamdey: about naming thing, I like when it is one-line self-explanatory:

I'm not sure about `messageParts`, it looks like it the request object from axios but not sure. Actually I don't get why `req` becomes `messageParts`.

And  `priv` is a but too generic imo

MonsieurLanza:
No, this commit was really about getting things work in the worker. I'm on the body right now.

MonsieurLanza:
Known issues & gotchas : 
- some polyfills may be missing (especially TextEncoder),
- `puleos/object-hash` insert object's type before hashing, we may want to replace that lib,
- I really need to add tests,
- signup and import contact requests are not signed yet.

This is still WIP, and not ready to be merged. Besides this, it works.

src/frontend/web_application/src/services/api-client/index.js
```diff
@@ -24,16 +28,54 @@ if (BUILD_TARGET === 'server') {
...
+    headers,
+    paramsSerializer: params => queryStringify(params, headers),
+    transformRequest: ([(data) => {
+      if (data instanceof Blob) {
+        if (typeof FormData !== 'function') {
+          throw new Error('not a browser environment');
+        }
+
+        const transformedData = new FormData();
+        transformedData.append('attachment', data);
```
iamdey: I'm not sure it is a good idea to put here the attachment (the term is related to message's attachment). 

Importing contacts might not be named the same.

MonsieurLanza: I put it in a specific object, to be able to carry the field name with it all the way to the transformer. Object which also handle transformation to formdata and the browser env check that comes with it.

PR #932, opened 2018-05-17
Feature/backend/resync index

gdchamal:
Need to be able to use a different mapping version than the current one as current command does

### DIFF
--- a/src/backend/tools/py.CLI/caliopen_cli/commands/resync_index.py
+++ b/src/backend/tools/py.CLI/caliopen_cli/commands/resync_index.py
@@ -11,8 +11,8 @@ import sys
 from elasticsearch import Elasticsearch
 from caliopen_storage.config import Configuration
 
-log = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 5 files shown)


src/backend/tools/py.CLI/caliopen_cli/commands/resync_index.py
```diff
@@ -24,21 +24,33 @@ def resync_index(user_name, **kwargs):
...
 
     es_url = Configuration('global').get('elasticsearch.url')
     es_client = Elasticsearch(es_url)
 
     # Delete current index
-    if not es_client.indices.exists([index]):
-        log.warn('Index not found for user %s' % user_name)
+    if not es_client.indices.exists([old_index]):
+        log.warn('Index %r not found for user %s' % (old_index, user.name))
         sys.exit(1)
```
sapiens-sapide: why are you exiting ? You may want to create an index if it is missing, won't you ?

gdchamal: it's not a normal situation, may index is in a wrong version etc. I prefer to deal with that situation manually

PR #935, opened 2018-05-18
Disable button "Copy To" in reply form

### LINKED ISSUE #903: wrong command: "copy to" saves a draft
When I write a message in a discussion, _copy to_ saves a draft instead of copying the message.
As I understand it, the copy function is not available yet, so the button should be greyed out and noninteractive? Instead it is producing the wrong command.

capture d ecran de 2018-05-03 15-06-04

MonsieurLanza:
Fixes #903

### DIFF
--- a/src/frontend/web_application/src/modules/draftMessage/components/ReplyForm/presenter.jsx
+++ b/src/frontend/web_application/src/modules/draftMessage/components/ReplyForm/presenter.jsx
@@ -174,16 +174,23 @@ class ReplyForm extends Component {
             >
               <Trans id="messages.compose.action.save">Save</Trans>
             </Button>
-            <Button className="m-reply__bottom-action" onClick={this.handleSave} icon="share-alt" responsive="icon-only">
-              <Trans id="messages.compose.action.copy">Copy to</Trans>
-            </Button>
-            {/* TODO: enable rich text editor
-              <Button
-              className="m-new-reply__bottom-action m-reply__bottom-action--editor"
-              icon="editor"
-              responsive="icon-only"
-            />
-            */}
+            {
+            // XXX no api available yet
+            // <Button
+            //   className="m-reply__bottom-action"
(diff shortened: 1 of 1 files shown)


src/frontend/web_application/src/modules/draftMessage/components/ReplyForm/presenter.jsx
```diff
@@ -174,7 +174,12 @@ class ReplyForm extends Component {
             >
               <Trans id="messages.compose.action.save">Save</Trans>
             </Button>
-            <Button className="m-reply__bottom-action" onClick={this.handleSave} icon="share-alt" responsive="icon-only">
+            <Button
+              className="m-reply__bottom-action"
+              icon="share-alt"
+              responsive="icon-only"
+              disabled
```
iamdey: I prefer commented instead of disabled.

Or notify "not yet implemented" is legit  too.

src/frontend/web_application/src/modules/draftMessage/components/ReplyForm/presenter.jsx
```diff
@@ -174,7 +174,11 @@ class ReplyForm extends Component {
             >
               <Trans id="messages.compose.action.save">Save</Trans>
             </Button>
-            <Button className="m-reply__bottom-action" onClick={this.handleSave} icon="share-alt" responsive="icon-only">
+            // <Button
+            //   className="m-reply__bottom-action"
+            //   icon="share-alt"
+            //   responsive="icon-only"
+            // >
               <Trans id="messages.compose.action.copy">Copy to</Trans>
```
iamdey: these 2 lines should be commented too

iamdey: can you add a comment like `// XXX no api available yet` or so

PR #943, opened 2018-05-24
add option to enable syslog

iamdey:
syslog is disabled by default

start with the following (or in conf file) to enable it:

```
CALIOPEN_ENABLE_SYSLOG=true bin/server
```


iamdey:
So I merge this since the current docker build is broken

PR #953, opened 2018-05-26
loose configuration for docker containers

iamdey:
+ allow to stop one container and replace by native app
+ fix port exposure for apiv1

### DIFF
--- a/devtools/docker-compose.yml
+++ b/devtools/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '2.1'
 volumes:
   index:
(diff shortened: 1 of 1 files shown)


devtools/docker-compose.yml
```diff
@@ -1,22 +1,25 @@
-version: '2'
+version: '2.1'
```
iamdey: not sure this is useful

PabloPie: https://docs.docker.com/compose/compose-file/compose-versioning/#version-21

PR #955, opened 2018-05-27
Add the possibility to build an image

iamdey:
update the doc frontend install, but almost ready for any kind of developments


PabloPie:
LGTM, this way if you pull the image before building you don't have to build every layer, may save some time...

iamdey:
we only need some doc to explain how to publish a new images for granted people ;)

PR #967, opened 2018-06-01
Fix/frontend/discussion sort

iamdey:


iamdey:
actually the days grouping doesn't work well

iamdey:
should be ok now

PR #971, opened 2018-06-08
Feature/backend/go cli

sapiens-sapide:
Add a new general purpose CLI to stack : `gocaliopen`
Included in this PR command `fixMissingParticipants` that try to fill-up participants property in messages that miss it into cassandra.  
Run `gocaliopen fixMissingParticipants` for help.


src/backend/tools/go.CLI/cmd/gocaliopen/cli_cmds/fixMissingParticipants.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		log.Infoln("failed to retrieve participant from index, trying to reinject raw message")
+		err = reInjectRaw(usrId.String(), rawMsgId.String(), MsgQueue)
+		if err != nil {
+			log.WithError(err).Warn("failed to re-inject raw message in stack")
+			failedCount++
+			return
+		}
+		//reinjection OK, delete former invalid message
+		err = Store.Session.Query(`DELETE FROM message WHERE message_id = ? AND user_id = ?`, msgId, usrId).Exec()
```
gdchamal: there is some state on previous message that we will make disappear: read status at least

src/backend/tools/go.CLI/cmd/gocaliopen/cli_cmds/root.go
```diff
@@ -0,0 +1,272 @@
...
+	apiCfg := viper.New()
+	lmtpCfg := viper.New()
+
+	if cfgPath != "" {
+		// Use path from the flag
+		apiCfg.AddConfigPath(cfgPath)
+		lmtpCfg.AddConfigPath(cfgPath)
+	} else {
+		// set defaults to relative path from /tools/go.CLI/cmd/gocaliopen
+		apiCfg.AddConfigPath("../../../../configs")
```
gdchamal: I still really don't like this relative path usage that are really uncommon. Can we remove such default values. Yes we will have to provide configPath or have a correct environment, that's is better than this usage.

PR #972, opened 2018-06-12
refactor request & create draft for reply and compose

### LINKED ISSUE #894: "reply" button nonfunctional
While in a discussion, hitting _reply_ on a message doesn't do anything. 
_#reply_ get added at the end of the adress bar, but nothing happen inside the page.
I believe the page should scroll down to the reply box, with the selected message in quote?

capture du 2018-05-03 12-42-34

linked to #521

iamdey:
fix #894 button reply has no effects


src/frontend/web_application/src/modules/message/actions/requestMessages.js
```diff
@@ -0,0 +1,6 @@
+import { requestMessages as requestMessagesBase } from '../../../store/modules/message';
+import { tryCatchAxiosAction } from '../../../services/api-client';
+
+// @deprecated: use fetchMessages instead or request collection directly
```
MonsieurLanza: Why not refactor requestMessages directly ? Does it break many things ?

iamdey: yes :)

PR #977, opened 2018-06-14
POC: flow type

iamdey:
use comments to declare types
It would probably be necessary to set weak mode


iamdey:
flow

iamdey:
WDYT ? go or not go ?

other solutions : use inline typing (not in comments)
maybe in specific files (didn't find how for now) something like `src/lib/foo.js`  & `src/lib/foo.js.flow` for example
...

PR #982, opened 2018-06-20
Update remote identities' credentials without providing previous ones.

sapiens-sapide:


sapiens-sapide:
[Gestion des remote identities](https://trello.com/c/5hTPTx29/86-gestion-des-remote-identities)

PR #984, opened 2018-06-21
Feature/backend/notify remote fetch failures

sapiens-sapide:
API will set remote_identity.status=inactive if fetching failed for more than 48hours. (without emitting a notification for now).


sapiens-sapide:
[Gestion des remote identities](https://trello.com/c/5hTPTx29/86-gestion-des-remote-identities)

PR #987, opened 2018-06-26
Fix discussion «load more»

MonsieurLanza:
As for now, fixes incorrect behavior on "load more" in discussions : new messages were appended to bottom of page.

### DIFF
--- a/src/frontend/web_application/src/scenes/MessageList/index.js
+++ b/src/frontend/web_application/src/scenes/MessageList/index.js
@@ -25,7 +25,7 @@ const mapStateToProps = createSelector(
   (messagesById, discussionId, {
     didInvalidate, messageIds, hasMore, isFetching,
   }) => {
-    const messages = messageIds.map(messageId => messagesById[messageId]);
+    const messages = messageIds.map(messageId => messagesById[messageId]).reverse();
 
     return {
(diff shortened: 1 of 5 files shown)


src/frontend/web_application/src/scenes/MessageList/index.js
```diff
@@ -25,7 +25,7 @@ const mapStateToProps = createSelector(
   (messagesById, discussionId, {
     didInvalidate, messageIds, hasMore, isFetching,
   }) => {
-    const messages = messageIds.map(messageId => messagesById[messageId]);
+    const messages = messageIds.map(messageId => messagesById[messageId]).reverse();
```
iamdey:
```
// use date_sort property to sort correctly messages
const messages = messageIds.map(messageId => messagesById[messageId]).sortByDateSort().reverse();
```

MonsieurLanza:
This PR is for a feature that will be of little use once timeline will be «discussion based» rather than «message based».
That said, the fixes already done are useful, maybe re-target it ?

src/frontend/web_application/src/scenes/MessageList/index.js
MonsieurLanza: Fixed

PR #993, opened 2018-07-02
fix bad props overwriting for remote.info

iamdey:
this caused identity to be polled as new every time an update was done

+ fix bad name to get the active value. previsouly a new remote wasn't enabled by default


iamdey:
tests are failing again due to call to action on tag management

PR #994, opened 2018-07-03
update functionnal tests

iamdey:
still not perfect because tags specs works when previous tests about adding/removing messages is done.

Will be better when the action bar will be implemented


iamdey:
ok so we'll wait for it to be merged

iamdey:
I close since a lot of things will change with next ui

PR #1007, opened 2018-07-13
wait for the signin form being mounted to prevent not xhr requests on…

### LINKED ISSUE #1001: fake device
https://feedback.caliopen.org/t/bug-du-a-un-changement-interface-et-nouvelles-implementations/526

iamdey:
… signin

fix #1001

And replace safe, public and unsafe login buttons by simple login button


iamdey:
tests fail because not full up to date and not clickable elements

PR #1010, opened 2018-07-13
Fix/backend/imap worker err

sapiens-sapide:


gdchamal:
I discard this PR it was based on develop but must be on master. Changing branch, rebasing is not enough as it take commits from develop, I'll replay commit on a new branch

PR #1014, opened 2018-07-16
refactor sass bootstrap

iamdey:
+ bootstrap correctly foundation
+ prevent duplication of reset styles (optim)
+ remove unecessary foundation css classes


src/frontend/web_application/webpack/config.js
```diff
@@ -18,7 +16,7 @@ const configureStylesheet = () => {
       }),
       // FIXME: useful for deduplication, we need a way prevent that.
       // disabled for dev for build time (10sec instead of 30+)
-      ...(!isDev ? [new OptimizeCssAssetsPlugin({ canPrint: false })] : []),
+      new OptimizeCssAssetsPlugin({ canPrint: false }),
```
MonsieurLanza: It is no more disabled for dev build, or is it ? What about deduplication ?

src/frontend/web_application/src/App.jsx
```diff
@@ -9,6 +9,7 @@ import Routes from './routes';
 import I18nProviderLegacy from './components/I18nProvider';
 import { initConfig } from './services/config';
 import { NotificationProvider } from './modules/notification';
+import './app.scss';
```
MonsieurLanza: :+1:

src/frontend/web_application/webpack/config.js
iamdey: it is no more useful thanks to the fix of bootstrap foundation:

previously each time we required it (for flex-grid for example) all the foundation classes were duplicated and so with optimize plugin, the build time was very long and without the plugin the build time was just long.

now with or without, it is almost the same

But without I had some strange behaviors (width of important slider)

iamdey: oh. I forgot to remove the comment

PR #1018, opened 2018-07-19
Feature/backend/discussion v2

gdchamal:
Enhance discussion objects returned by API

Try to build discussions by participants


gdchamal:
[Gestion discussion](https://trello.com/c/Jyjxnib5/117-gestion-discussion)

sapiens-sapide:
Let's try with this, but we may facing two issues : 
- we'll fail to connect a participant to the relevant discussion if (s)he has multiple addresses (need to link to a contact)
- will coexistence of two models of discussions (legacy and this new one) be usable ? Or do we need to rebuild discussions from scratch on alpha ?

PR #1025, opened 2018-07-26
Fix/backend/apiv1/no fake device

gdchamal:
Remove fakedevice usage, on authentications we strictly need device informations.

Remove constraint about only one connection per user, now one connection is possible per device

### DIFF
--- a/src/backend/interfaces/REST/py.server/caliopen_api/user/user.py
+++ b/src/backend/interfaces/REST/py.server/caliopen_api/user/user.py
@@ -13,7 +13,7 @@ from .util import create_token
 
 from ..base import Api
 from ..base.exception import AuthenticationError, NotAcceptable
-from ..base.exception import Unprocessable, ValidationError
+from ..base.exception import Unprocessable, ValidationError, MethodFailure
 
 from caliopen_storage.exception import NotFound
 from caliopen_main.common.core import PublicKey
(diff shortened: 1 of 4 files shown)


gdchamal:
[Remove fake device usage](https://trello.com/c/Xs0AB9jJ/121-remove-fake-device-usage)

src/backend/interfaces/REST/py.server/caliopen_api/user/user.py
```diff
@@ -125,9 +125,6 @@ def collection_post(self):
                                  'curve': key.crv})
         self.request.cache.set(cache_key, session_data)
```
sapiens-sapide: what happen if `key = None ` ?

gdchamal: it crash later as application try to find a curve and it's not OK (device exist but public key was not created when creating the device)

I don't know which error code to raise, instead of only log an error when key is not found. Any idea ?

sapiens-sapide: something similar to `raise ValidationError(detail='No device informations')` ?

gdchamal: It's not a client error, so use of a 4xx http error is not the best solution and can't find any 5xx error that could fit (or a generic 500 ?).

sapiens-sapide: 424 ? (event if it is not client error, it is dependency fail, isn't it ?)

PR #1026, opened 2018-07-26
how to fix docker won't build backend services

iamdey:


doc/install/frontend-development.md
```diff
@@ -157,3 +157,18 @@ or rebuild node-sass:
...
 ```
+
+**docker container can't build or start**
+
+the following error can appears when the host user (you) can use directly  docker and when a file is owned by root only:
+
+> Building api
+> ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
+
+It is mostly due to minio which updates its config file without asking.
```
MonsieurLanza: Not sure about "mostly"

doc/install/frontend-development.md
```diff
@@ -157,3 +157,18 @@ or rebuild node-sass:
 ```
 npm rebuild node-sass
 ```
+
+**docker container can't build or start**
+
+the following error can appears when the host user (you) can use directly  docker and when a file is owned by root only:
```
MonsieurLanza: **T**he following… (missing cap)

doc/install/frontend-development.md
```diff
@@ -157,3 +157,18 @@ or rebuild node-sass:
...
+
+> Building api
+> ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
+
+It is mostly due to minio which updates its config file without asking.
+
+You need to change the owner of the file (please don't commit) until we find a solution.
```
MonsieurLanza: How ?

doc/install/frontend-development.md
PabloPie: "can appear" also, and the next phrase is not very explicit, I would say "when docker is not run as root"

doc/install/frontend-development.md
PabloPie: Not needed IMO.

iamdey:
@gdchamal I think we need a new pass on the doc about docker and eventually minikube when stable

PabloPie:
Sure, the thing is it maybe won't be needed as image building and publishing will be automated.

PR #1040, opened 2018-09-03
go back to the previous tab if exist when closing a tab

iamdey:
+ refactor TabProvider to not use redux for location


iamdey:
one edge case: when previous location is the same tab (settings), it goes back to home instead of real previous tab

PR #1015, opened 2018-07-18
WIP: refactor Page layout & Header

iamdey:

PR #1041, opened 2018-09-04
[WIP] global actions on discussion/timeline + reply

MonsieurLanza:


src/frontend/web_application/src/scenes/Discussion/components/MessagePi/index.jsx
```diff
@@ -88,21 +88,23 @@ class MessagePi extends PureComponent {
...
+            aria-valuemax="100"
+            tabIndex="0"
+          >
+            <div
+              className={classnames('m-message-pi__progress-meter', `m-message-pi__progress-meter--${getPiClass(piAggregate)}`)}
+              style={{ width: `${piAggregate}%` }}
+            />
+          </div>
+          <div className="m-message-pi__numeric">
+            <span className="m-message-pi__numeric--legend">Privacy index&thinsp;:</span>
```
iamdey: `legend` & `value` are not modifiers like describe in BEM notation, it should be named:

```
m-message-pi__numeric-legend
m-message-pi__numeric-value
```

src/frontend/web_application/src/scenes/Discussion/components/MessagePi/style.scss
```diff
@@ -76,21 +86,30 @@
...
+
+  @include breakpoint(medium) {
+    &__illustration, &__meter, &__numeric {
+      width: 100%;
+    }
+
+    &__metaphor {
+      padding: 0.825rem 0;
+      border-bottom: 2px solid #999999;
+      font-size: 0.625rem;
```
iamdey: As discussed before summer, we should leave the responsibility of the font size to the body: https://github.com/CaliOpen/Caliopen/pull/1015

src/frontend/web_application/src/scenes/Discussion/components/MailMessage/index.jsx
```diff
@@ -81,76 +58,54 @@ class MailMessage extends Component {
...
+      .map(participant => participant.label).join(', ');
 
+    return `${prefix}${otherRecipients}`;
+  }
+
+  renderTags = ({ tags }) => (
+    tags && (
+      <ul className="s-mail-message__tags">
+        {tags.map(tag => (
+          <li key={`${this.props.message.message_id}${tag}`} className="s-mail-message__tag"><Badge>{tag}</Badge></li>
```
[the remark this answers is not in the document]
MonsieurLanza: > I'm not sure but I think a unique id is not mandatory this will not be on the same node tree.
It did complain about non uniqueness, this is-why I added message id

src/frontend/web_application/src/scenes/Discussion/components/MessagePi/style.scss
MonsieurLanza: According to UI mock-up, this should be smaller than body font, and is relative to it, not sure about what you mean here ? Too small ?

PR #1048, opened 2018-09-11
Feature/backend/proxy api

gdchamal:
Apiv1 and apiv2 are proxyfied using another container with a nginx.

Apiv1 run in its container behind a nginx and using uwsgi workers to fit a production setup


src/backend/Dockerfile.py-api
```diff
@@ -6,6 +6,15 @@
...
 MAINTAINER Caliopen
 
+RUN apk add nginx linux-headers supervisor
+RUN pip install --upgrade pip
+RUN pip install uwsgi
+
+# Copy configuration files to run apiv1 behind uwsi and nginx
+COPY ./configs/apiv1-nginx.conf /etc/nginx/nginx.conf
+COPY ./configs/apiv1-uwsgi.ini /etc/uwsgi/uwsgi.ini
+COPY ./configs/apiv1-supervisord.conf /etc/supervisor.d/apiv1.ini
```
PabloPie: Can't we mount these like we do with the rest of the configuration files?

gdchamal: In a previous version it's what I did, but I think it's better to build the image this way. We can mount another version of one of this file in the cluster that will overwrite this version for 'development'.

Doing this way the image is more independant than the infrastructure described in the compose file.

PabloPie:
lgtm

sapiens-sapide:
We could have set "withProxy" option to `false` by default in go.server to avoid removing proxy code. It'd have the benefit of keeping a simple and useful proxy for development ! :)

gdchamal:
I revert my commits removing the proxy, but instead I set the default value to false to not activate it

PR #1059, opened 2018-09-28
UIv4 Corrections to prepare merge

MonsieurLanza:
- Comment out unavailable batch operations
- Fix Reply on Message (still an error in middleware)
- Some CSS corrections/refactor.

### DIFF
--- a/src/frontend/web_application/src/scenes/Discussion/components/MailMessage/style.scss
+++ b/src/frontend/web_application/src/scenes/Discussion/components/MailMessage/style.scss
@@ -26,14 +26,6 @@
     width: 100%;
     padding: 1rem;
 
-    &-from, &-to {
-      font-size: 0.625rem;
-
-      a, a:visited {
-        color: #515151;
-        font-weight: 600;
-      }
-    }
   }
 
 
@@ -48,10 +40,6 @@
 
   &__participants, &__tags {
     display: none;
-  }
-
-  .direction {
-    width: 2rem;
   }
 
   &__details {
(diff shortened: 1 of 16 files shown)


src/frontend/web_application/src/scenes/Discussion/components/MailMessage/style.scss
```diff
@@ -134,6 +122,17 @@
       display: block;
     }
 
+    &__participants-from, &__participants-to {
+      overflow: hidden;
+      text-overflow: ellipsis;
+      font-size: $co-font__size--small;
+
+      a, a:visited {
```
iamdey: that's still not a good.

src/frontend/web_application/src/scenes/Discussion/style.scss
```diff
@@ -14,26 +14,36 @@
     background-color: white;
 
     .s-discussion__header {
-      padding: 0 2.625rem;
+      padding: 0 40px;
```
iamdey: hu ?!

src/frontend/web_application/src/scenes/Discussion/style.scss
```diff
@@ -14,26 +14,36 @@
...
+    flex-grow: 1;
+    align-items: center;
+    justify-content: flex-end;
+  }
+
   &__action-label {
     @include breakpoint(small) {
       display: none;
     }
+    display: block;
```
iamdey: it should be the default value unless it is an inline element 
(duno how to deal with this kind of behavior: it is mandatory and so DOM can change without breaking anything or the DOM element was not good at the beginning).

PR #1064, opened 2018-10-04
Some polish on PI display on messages

MonsieurLanza:
Changed class/related names for pi qualities (ugly, bad, good, super). Use theme palettes whenever possible.
Still needs.
- [x] texts/translations for PI metaphor


iamdey:
I don't get why change the names of pi qualities.

MonsieurLanza:
I did this because there was more than 2 levels in already existing privacy palettes, and for consistency naming across all UI. 

So I reused the 4 names of existing palette.

PR #1075, opened 2018-10-18
Chore/frontend/update lingui

iamdey:


src/frontend/web_application/src/scenes/ApplicationSettings/components/NotificationSettings/presenter.jsx
```diff
@@ -40,7 +40,7 @@ class NotificationForm extends Component {
     const messagePreviewOptions = this.getOptionsFromArray(MESSAGE_PREVIEW);
     const delayDisappearOptions = DELAY_DISAPPEAR.map(delay => ({
       value: delay,
-      label: i18n._('settings.notification.delay_disappear.options.second', null, { values: { 0: delay }, defaults: '{0} Seconds' }),
+      label: i18n._('settings.notification.delay_disappear.options.second', { 0: delay }, { defaults: '{0} Seconds' }),
```
iamdey: but sometimes it looks like it works with objects too :s

src/frontend/web_application/src/scenes/DevicesSettings/presenter.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { Trans } from '@lingui/react';
+import { Trans } from '@lingui/macro'; // eslint-disable-line import/no-extraneous-dependencies
```
iamdey: very important!

when a `Trans` includes a component there is 2 possibilities: 
* declare all the things
* use `macro` a dev dependencies instead that will transpil into a full declaration including the importation itself:

```
// src
import { Trans } from '@lingui/macro';
//...
<Trans id="foo"><p>Hello<br />{name}</p></Trans>
```

```
// semi-transpiled
import { Trans } from '@lingui/react';

//...
<Trans id="foo" components={[<p/>, <br />]} values={{ name: 'foo' }} defaults="<0>Hello<1 />{name}</0>" />
```

without macro we must use semi-transpiled format

cf. https://lingui.js.org/ref/macro.html#jsx-macros

iamdey:
all functional tests are ok on local env

PR #1089, opened 2018-11-09
CRUD on public keys

MonsieurLanza:
Misses styling & component redraw.


src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
```diff
@@ -0,0 +1,145 @@
...
+const TextareaFieldGroup = renderReduxField(TextareaFieldGroupBase);
+
+@withI18n()
+class PublicKeyForm extends PureComponent {
+  static propTypes = {
+    errors: PropTypes.arrayOf(PropTypes.string),
+    i18n: PropTypes.shape({}).isRequired,
+    contactId: PropTypes.string.isRequired,
+    publicKey: PropTypes.shape({}),
+    handleSubmit: PropTypes.func.isRequired,
```
iamdey: handleSubmit or onSubmit ?

my notation rule is:
- if it is a component prop: `onWhatever`
- if it is a class method: `handleWhatever`

for example:

```
class Foo extends Component {
  handleBleep = ({ bleep }) => { 
    // do things with bleep
  }
  
  render() {
    return <Bar onBleep={this.handleBleep} />
  }
}
```

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
```diff
@@ -0,0 +1,145 @@
...
+            contactId,
+            publicKey: { ...values, publicKeyId },
+            original: { label },
+          }));
+      } else {
+        handleSubmit(ev)
+          .then(values => createPublicKey({ contactId, publicKey: values }));
+      }
+    } catch (ex) {
+      return;
```
iamdey: may be notify something went wrong or display error.

If it is used to prevent an unpredictable error, I prefer to not fail silently but crash with our magnificent BSOD.

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
```diff
@@ -0,0 +1,145 @@
...
+    errors: PropTypes.arrayOf(PropTypes.string),
+    i18n: PropTypes.shape({}).isRequired,
+    contactId: PropTypes.string.isRequired,
+    publicKey: PropTypes.shape({}),
+    handleSubmit: PropTypes.func.isRequired,
+    form: PropTypes.string.isRequired,
+    createPublicKey: PropTypes.func.isRequired,
+    updatePublicKey: PropTypes.func.isRequired,
+    deletePublicKey: PropTypes.func.isRequired,
+    onSuccess: PropTypes.func,
```
iamdey: it is not an agnostic lib so we can required it because we are sure the parent needs this
(but I might be wrong if your some cases when it is not required)

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
```diff
@@ -0,0 +1,145 @@
...
+    onCancel: PropTypes.func,
+  };
+
+  static defaultProps = {
+    publicKey: undefined,
+    onSuccess: undefined,
```
iamdey: usually I set a noop: `onSuccess: () => {},`

so you will not have to add a condition afterward: `if (onSuccess) onSuccess()`

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
```diff
@@ -0,0 +1,145 @@
...
+                </Legend>
+              </FormColumn>
+              {errors.length > 0 && <FormColumn><FieldErrors errors={errors} /></FormColumn>}
+            </FormRow>
+            <FormRow>
+              <FormColumn rightSpaced={false}>
+                <Field
+                  component={TextFieldGroup}
+                  name="label"
+                  label={i18n._('contact.public_key_form.label.label', { defaults: 'Key label' })}
```
iamdey:
```
  label={i18n._('contact.public_key_form.label.label', null, { defaults: 'Key label' })}
``` 
?

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
```diff
@@ -0,0 +1,145 @@
...
+              </FormColumn>
+            </FormRow>
+            <FormRow>
+              <FormColumn rightSpaced={false}>
+                <Field
+                  component={TextareaFieldGroup}
+                  label={i18n._('contact.public_key_form.key.label', null, { defaults: 'Key (ascii armored)' })}
+                  name="key"
+                  required={publicKey === undefined}
+                  disabled={publicKey !== undefined}
```
iamdey: to save some digital ink:

```
required={!publicKey}
disabled={!!publicKey}
```

:D

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/style.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../../../styles/common';
+@import '../../../../styles/vendor/bootstrap_foundation-sites';
+
+.m-public-key-form {
+  padding-left: map-get($co-form__spacing, 'medium');
```
iamdey: `map_get` ?

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/style.scss
```diff
@@ -0,0 +1,15 @@
...
+  background-color: $co-color__bg__back--higher;
+
+  &__actions {
+    display: flex;
+  }
+
+  &__button-cancel {
+    margin-left:auto;
```
iamdey: we called this "ninja space", a missing or excess space (or tabs)

(this is obviously not important but sometimes very annoying when refactoring using regexp etc.)

src/frontend/web_application/src/scenes/Contact/components/PublicKeyList/presenter.jsx
```diff
@@ -0,0 +1,119 @@
...
+      publicKeys, contactId, requestPublicKeys,
+    } = this.props;
+
+    if (publicKeys.length === 0 || this.state.didInvalidate) {
+      requestPublicKeys({ contactId });
+    }
+  }
+
+  static getDerivedStateFromProps(props, state) {
+    return { ...state, didInvalidate: props.didInvalidate };
```
iamdey: I don't like this method: 

> This method exists for rare use cases

cf. https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

I translate it like this: "don't use this lifecycle method because it will break your application and it kills kitten."

A bad usage may result in an infinite render loop.

--- 

I might understand what you tried (I've made this error too):

I think you don't need `didInvalidate` in state at all.

```
componentDidMount() {
  if ((emptyPubKeys || this.props.didInvalidate) && !pubKeysRequestIsRunning) {
    requestPubKeys()
  }
}

// no componentDidUpdate()

onSuccess = () => {
  this.setState({ editMode: false }, () => {
    requestPubKeys();
  });
}
```

src/frontend/web_application/src/scenes/Contact/components/PublicKeyList/presenter.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+  enterEditMode = () => {
+    this.setState({ editMode: true });
+  }
+
+  quitEditMode = () => {
+    this.setState({ editMode: false });
+  }
+
+  handleEdit = publicKey => () => this.setState({ editMode: publicKey.key_id });
```
iamdey: `editMode` should not change its type

src/frontend/web_application/src/scenes/Contact/components/PublicKeyList/presenter.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+  render() {
+    const { publicKeys, contactId } = this.props;
+
+    return (
+      <ul>
+        {publicKeys.map(publicKey => (
+          this.state.editMode === publicKey.key_id ?
```
iamdey: please avoir  ternary operation for jsx
use sub functions for readability

src/frontend/web_application/src/scenes/Contact/components/PublicKeyList/presenter.jsx
```diff
@@ -0,0 +1,119 @@
...
+              <Icon
+                type={KEY_QUALITY_ICONS[this.getKeyQuality(publicKey)]}
+                className={`m-public-key-list__quality-icon--${KEY_QUALITY_CLASSES[this.getKeyQuality(publicKey)]}`}
+                rightSpaced
+              />
+              <strong className="m-public-key-list__key-label">{publicKey.label}</strong>&nbsp;:&nbsp;{publicKey.fingerprint}
+              <Button icon="edit" className="m-public-key-list__edit-button" onClick={this.handleEdit(publicKey)} />
+            </li>
+          )))}
+        { this.state.editMode === true ?
```
iamdey: well I don't get, it should throws on jslint & at react runtime; not ternary please

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
MonsieurLanza: This is kind of a reflex : a component might be reused elsewhere in the project, in a slightly different context. I really don't like to tie a component to its parent if I don't need to. At its root, the very concept of component is re-usability.

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/index.js
```diff
@@ -0,0 +1,25 @@
...
+});
+
+const mapDispatchToProps = dispatch => ({
+  ...bindActionCreators({
+    requestPublicKeys,
+    createPublicKey,
+    updatePublicKey,
+    deletePublicKey,
+  }, dispatch),
+  onSubmit: values => Promise.resolve(values),
```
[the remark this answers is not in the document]
MonsieurLanza: To be honest, I just used the same pattern as you did here : https://github.com/CaliOpen/Caliopen/blob/2a7b076703e35fb81aa702d57e72a9464d9547c5/src/frontend/web_application/src/scenes/Contact/index.js#L45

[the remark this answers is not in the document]
iamdey: I was young and naive :D

src/frontend/web_application/src/scenes/Contact/components/PublicKeyForm/presenter.jsx
iamdey: I understand what you mean, I have this reflex too; but like this case it is almost 90% sure this kind of components will not be reused and it will be more efficient to be straight forward and keep it very specific.

Agree too to say it is not big deal here.
I only want to prevent hours of debugging because for example someone removed or made a typo in the handle on `onsuccess` and get stucked on "Hey, why this doesn't work no more?!" without any warning.

src/frontend/web_application/locale/fr/messages.json
```diff
@@ -323,6 +332,7 @@
   "passwords.form.error.empty": "Au moins un champ est requis",
   "passwords.form.error.identifiants_mismatch": "Les identifiants ne correspondent pas.",
   "passwords.form.error.user_not_found": "Utilisateur introuvable.",
+  "public-keys-list.add-key.label": "",
```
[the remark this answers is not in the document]
MonsieurLanza: No, this key is unused, replaced by `contact.public_keys_list.add_key.label`. Deleted it.

src/frontend/web_application/src/scenes/Contact/components/PublicKeyList/presenter.jsx
```diff
@@ -64,54 +66,76 @@ class PublicKeyList extends Component {
...
-                rightSpaced
-              />
-              <strong className="m-public-key-list__key-label">{publicKey.label}</strong>&nbsp;:&nbsp;{publicKey.fingerprint}
-              <Button icon="edit" className="m-public-key-list__edit-button" onClick={this.handleEdit(publicKey)} />
-            </li>
-          )))}
-        { this.state.editMode === true ?
+    if (this.state.editMode && this.state.editKey === publicKey.key_id) {
+      return (
+        <li>
```
iamdey: no key needed here ?

PR #1093, opened 2018-11-21
disable reply when twitter DM is parent_message

iamdey:
+ fix locale catalog


src/frontend/web_application/src/modules/draftMessage/components/ReplyForm/presenter.jsx
```diff
@@ -99,94 +101,102 @@ class ReplyForm extends Component {
...
-              <Link to={`#${parentMessage.message_id}`} className="m-reply__parent-link">
-                <Trans id="reply-form.in-reply-to">In reply to: {excerpt}</Trans>
-              </Link>
-            </div>
-          )}
-          <BodyRow className="m-reply__body">
-            <DiscussionTextarea
-              body={this.state.draft.body}
-              onChange={this.handleChange}
+    if ((parentMessage && parentMessage.type === 'email') || (lastMessage && lastMessage.type === 'email')) {
```
MonsieurLanza: It's convenient that Twitter messages does not have a reply button ^^

iamdey: yes :)

PR #1110, opened 2018-11-23
remove caliopen_go service

iamdey:
caliopen_go actually not a service but a base image for go containers
it is already build by drone and pulled by docker directly on registry

+ remove "image" property. the only purpose was to simplified image name build on dev environment
this can be misunderstood because the property is mainly used to pull latest images from docker hub
also this might bring some conflicts in the eventual publish of Caliopen images on docker hub

### DIFF
--- a/devtools/docker-compose.yml
+++ b/devtools/docker-compose.yml
@@ -8,12 +8,6 @@ volumes:
     driver: local
 
 services:
-  caliopen_go:
-    image: public-registry.caliopen.org/caliopen_go
-    build:
-      context: ../src/backend
-      dockerfile: Dockerfile.caliopen-go
-
   # Proxy API
   proxy-api:
     image: nginx
@@ -133,7 +127,7 @@ services:
       - elasticsearch:elasticsearch
       - object-store:minio
       - nats:nats
-      - inbucket:inbucket
+      - inbucket:smtp
     ports:
       - "2525:2525"
     volumes:


iamdey:
I'm not 100% sure changing the image name will not break something. For now a basic usage docker-compose will have no impact.

devtools/docker-compose.yml
```diff
@@ -177,34 +166,33 @@ services:
...
       dockerfile: Dockerfile.imap-worker
-    image: caliopen_imap_worker
     links:
       - cassandra:cassandra
       - object-store:minio
       - elasticsearch:elasticsearch
       - nats:nats
     volumes:
       - ../src/backend/configs:/etc/caliopen
+      - /etc/ssl/certs:/etc/ssl/certs
```
PabloPie: This is a temporary workaround and should not be pushed, we risk never putting in place a real solution. Containers that need certificates need to include the certs.

iamdey:
also the certs are mounted from the host and might not work according to the distribution (not so much right ?)

devtools/docker-compose.yml
```diff
@@ -177,34 +166,33 @@ services:
     build:
       context: ../src/backend
       dockerfile: Dockerfile.imap-worker
-    image: caliopen_imap_worker
```
PabloPie: There isn't anything that should break with the change, I just like having images built with a more decent name.

Issue #1101, opened 2018-11-22
twitter worker lost connection to cassandra and don't reconnect


gdchamal:
On poc platform, after a certain time running, the twitter worker seem to have lost connection to cassandra and doesn't reconnect correctly:

```
twitter_worker_1_e8ae86a983fa | time="2018-11-22T14:46:18Z" level=info msg="[DMmsgHandler] ordering to pollDM for remote 2b59887a-b833-4fbc-877e-6b6ada534527 (user 0baf612a-e43e-45f5-be95-278a87281ca1)"
twitter_worker_1_e8ae86a983fa | time="2018-11-22T14:46:18Z" level=warning msg="[TwitterWorker] PollDM\u00a0failed to retrieve infos map for account @gdchamal (remote 2b59887a-b833-4fbc-877e-6b6ada534527, user 0baf612a-e43e-45f5-be95-278a87281ca1)" error="session has been closed"
twitter_worker_1_e8ae86a983fa | time="2018-11-22T14:46:18Z" level=warning msg="[TwitterWorker] ProcessInDM failed for event : {CreatedAt:1542895286960 ID:1065606184135286788 Type:message_create Message:0xc4202f41c0}" error="[Twitter Broker]SaveRawDM failed to store raw message in store : session has been closed"
twitter_worker_1_e8ae86a983fa | time="2018-11-22T14:48:18Z" level=info msg="received sync order for remote twitter ID 2b59887a-b833-4fbc-877e-6b6ada534527"
twitter_worker_1_e8ae86a983fa | time="2018-11-22T14:48:18Z" level=info msg="[DMmsgHandler] ordering to pollDM for remote 2b59887a-b833-4fbc-877e-6b6ada534527 (user 0baf612a-e43e-45f5-be95-278a87281ca1)"
twitter_worker_1_e8ae86a983fa | time="2018-11-22T14:48:18Z" level=warning msg="[TwitterWorker] PollDM\u00a0failed to retrieve infos map for account @gdchamal (remote 2b59887a-b833-4fbc-877e-6b6ada534527, user 0baf612a-e43e-45f5-be95-278a87281ca1)" error="session has been closed"
[...]

PR #1113, opened 2018-11-28
Fix/backend/polling issues

sapiens-sapide:
fixes and improvements. Should fix issue #1101  but I was unable to reproduce bug systematically…


sapiens-sapide:
Connection is not closed in this particular issue. Only `Session`.

PR #1127, opened 2018-12-05
Refactor of InstantMessage component

MonsieurLanza:
More BEM compliant, correct recipients and message from user restyle.

### DIFF
--- a/src/frontend/web_application/src/scenes/Discussion/components/InstantMessage/style.scss
+++ b/src/frontend/web_application/src/scenes/Discussion/components/InstantMessage/style.scss
@@ -12,7 +12,16 @@
   margin-bottom: 1rem;
 
   &--from-user {
-   flex-direction: row-reverse;
+    flex-direction: row-reverse;
+    margin-left: auto;
+
+    .m-instant-message__info {
+      background-color: #feffff;
+    }
+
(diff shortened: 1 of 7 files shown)


src/frontend/web_application/src/scenes/Discussion/components/InstantMessage/style.scss
```diff
@@ -28,76 +44,38 @@
...
-      background-color: #edccd4;
-    }
-
-    .progress-meter {
-      background-color: #ec0f48;
-    }
-
-    .pi-numeric {
-      color: #ec0f48;
+  .m-message-pi {
```
iamdey: not BEM, can you put it in a separate file, `message-pi.scss` for example ?

src/frontend/web_application/src/scenes/Discussion/components/MailMessage/style.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../../../styles/common';
 @import '../../../../styles/vendor/bootstrap_foundation-sites';
 
-.instant +.s-mail-message {
+.m-instant-message +.s-mail-message {
```
iamdey: not BEM, please add comment

PR #1145, opened 2018-12-12
Fix BSOD on quick reply.

MonsieurLanza:


src/frontend/web_application/src/modules/draftMessage/components/DraftMessage/presenter.jsx
```diff
@@ -153,7 +153,8 @@ class DraftMessage extends Component {
   getQuickInputPlaceholder = () => {
     const { i18n, draftMessage, availableIdentities } = this.props;
 
-    const [identityId] = draftMessage.user_identities;
+    const [identityId] = draftMessage && draftMessage.user_identities ?
+      draftMessage.user_identities : [];
```
iamdey: it's strange because a draft MUST have an identity.
So this means the draft might not be initialized (or worst half initialized) and so may be the problem should be fixed by reseting the state : cf render method: `this.renderPlaceholder();`
or may be using a key on the parent on this component for example `<DraftMessage key={draftMessage.message_id} />`

but well this is about the input placeholder, this fix could work as _dirty_, but I am not sure we will not have other cases like this

MonsieurLanza: Draft is sometimes `undefined`, like after a POST actions on message. As this is a quick fix I added `draftMessage.user_identities` in the condition just in case, but the main check is on `draftMessage` itself.

PR #1149, opened 2018-12-13
Feature/frontend/filter spam

iamdey:


src/frontend/web_application/src/scenes/Timeline/presenter.jsx
```diff
@@ -155,52 +175,78 @@ class Timeline extends Component {
     const { discussions, user, settings } = this.props;
     const { selectedDiscussions } = this.state;
 
-    if (discussions) {
+    if (!discussions.length) {
       return (
```
MonsieurLanza: I'd prefer the placeholder to be in its own render function.

Issue #1137, opened 2018-12-10
DM duplicated 


iamdey:
when DM are in "request" section in twitter, it appears twice on Caliopen. cf. https://mm.caliopen.org/caliopen/pl/63ka54h7xpbtzd91uae5kt1gme

I suppose it is an issue of twitter_worker, to be confirmed

PR #1151, opened 2018-12-14
Fix/backend/tweet duplicate

sapiens-sapide:
Prevent importing tweets twice by looking up external reference against tweets already fetched in database. Fix prevents sent tweet to appear twice and prevents re-importing tweets twice if user delete/create a Twitter account.
Should fix issue #1137

### DIFF
--- a/src/backend/main/go.backends/index/elasticsearch/messages.go
+++ b/src/backend/main/go.backends/index/elasticsearch/messages.go
@@ -33,6 +33,7 @@ func (es *ElasticSearchBackend) CreateMessage(user *objects.UserInfo, msg *objec
 		log.WithError(err).Warn("backend Index: IndexMessage operation failed")
 		return err
 	}
+
 	log.Infof("New msg indexed with id %s", resp.Id)
 	return nil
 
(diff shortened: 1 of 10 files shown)


src/backend/main/go.backends/index/elasticsearch/messages.go
```diff
@@ -174,6 +175,24 @@ func (es *ElasticSearchBackend) GetMessagesRange(filter objects.IndexSearch) (me
 	return messages, totalFound, nil
 }
 
+// return false if message not found or error
+func (es *ElasticSearchBackend) SeekMessageByExternalRef(userID, externalMessageID string) (bool, error) {
```
gdchamal: So an elasticsearch lookup for each message to check if it was previously received, I don't like it.

It should be best to create a lookup table in cassandra to map user_id, external_ref -> message_id

Hence no use of shard_id to query the index, is not ok

sapiens-sapide:
first two commits are from a previous fix because I rebased on fix branch before doing this one. They should be removed.

src/backend/main/go.backends/index/elasticsearch/messages.go
sapiens-sapide: OK. I close PR to make use of cassandra instead of elasticsearch and add a new lookup table : `message_externalref_lookup`

PR #1156, opened 2018-12-21
Feature/backend/tagger

maudetes:
Add tagging functionality to the develop branch.
To install the component, use --process-dependency-links, in order to install the right fasttext version (not available in pypi).
Additionally, to use the word_tokenizer in nltk, you need to download the appropriate models ("punkt"). 

Pre_processing might be done upstream instead.


src/backend/components/py.tag/caliopen_tag/models_manager/manager.py
```diff
@@ -0,0 +1,129 @@
...
+        file, following the fastText format.
+        Then it pre-processes the data (tokenization + lowercase).
+        Then it trains a model using fastText and saves it at output location.
+        Finally it removes the temporary file.
+        """
+        self._write_training_data_to_file()
+        self._train_tagging_model(output)
+        self._remove_tempfile()
+
+    def _write_usenet_data(self, f):
```
gdchamal: From my point of view these 2 methods (_write_usenet_data and _write_elastic_data) should not exists.

If to build the model we need 2 distincts data provider (usenet + elastic), we must combine them in a single DataProvider class.

My point is : give a single DataProvider instance to ModelManager.__init__ constructor, then the ModelManager will use this provider to iterate and format over items to build the model.

Then we can train different models with different DataProvider source

maudetes: I made some changes following this review. 

The DataManager takes a single DataProvider as constructor and simply iterates on it, as it is the general use case.

Additionally, the new MultipleSourceDataManager allows to make a wrapper of multiple providers and iterates on those successively.

maudetes: To learn a model combining data from Usenet and ES providers, we could do the following:
`
        usenet_manager = UsenetDataManager({})
        usenet_manager.prepare("/path/to/data.txt")

        config = Configuration('global').configuration
        es_manager = ESDataManager(config)
        query = es_manager.get_query()
        es_manager.prepare(query, index=None, doc_type='indexed_message')

        provider = MultipleSourceDataManager([usenet_manager, es_manager])

        for item in provider.next():
                print(item)
`

PR #1164, opened 2019-01-07
Feature/backend/send pgp mail

gdchamal:
This change permit to send correctly formed an encrypted mail using PGP method (RFC 3156)

The body will not be stored in index after message is marshalled into it's  mail envelope

### DIFF
--- a/src/backend/brokers/go.emails/email.go
+++ b/src/backend/brokers/go.emails/email.go
@@ -99,7 +99,7 @@ func (b *EmailBroker) MarshalEmail(msg *Message) (em *EmailMessage, err error) {
 	crypt_method, ok := features["message_encryption_method"]
 	if ok {
 		if crypt_method == "pgp" {
-			err := b.MarshalEncryptedEmail(msg, em, addr_fields)
+			err := b.MarshalPGPEmail(msg, em, addr_fields)
 			if err != nil {
 				return &EmailMessage{}, err
 			}
--- a/src/backend/brokers/go.emails/email.go
+++ b/src/backend/brokers/go.emails/email.go
@@ -52,6 +52,17 @@ func (b *EmailBroker) MarshalEmail(msg *Message) (em *EmailMessage, err error) {
 		},
 		Message: msg,
 	}
+	// sha256 internal message id to form external message id
+	hasher := sha256.New()
+	hasher.Write(em.Message.Message_id.Bytes())
+	sha := base64.URLEncoding.EncodeToString(hasher.Sum(nil))
+	messageId := sha + "@" + b.Config.PrimaryMailHost // should be the default domain in case there are multiple 'from' addresses
+
+	em.Message.Date = time.Now()
+
+	// Assign computed values
+	em.Message.Date_sort = em.Message.Date
+	em.Message.External_references.Message_id = messageId
 
 	m := gomail.NewMessage()
 	addr_fields := newAddressesFields()
@@ -82,18 +93,9 @@ func (b *EmailBroker) MarshalEmail(msg *Message) (em *EmailMessage, err error) {
 		}
 	}
 
-	em.Message.Date = time.Now()
-	em.Message.Date_sort = em.Message.Date
 	m.SetHeader("Date", em.Message.Date.Format(time.RFC1123Z))
-
-	// sha256 internal message id to form external message id
-	hasher := sha256.New()
-	hasher.Write(em.Message.Message_id.Bytes())
-	sha := base64.URLEncoding.EncodeToString(hasher.Sum(nil))
-	messageId := sha + "@" + b.Config.PrimaryMailHost // should be the default domain in case there are multiple 'from' addresses
 	m.SetHeader("Message-ID", "<"+messageId+">")
-
-	em.Message.External_references.Message_id = messageId
+	m.SetHeader("X-Mailer", "Caliopen-"+b.Config.AppVersion)
 
 	if msg.External_references.Parent_id != "" {
 		m.SetHeader("In-Reply-To", "<"+msg.External_references.Parent_id+">")
@@ -104,15 +106,12 @@ func (b *EmailBroker) MarshalEmail(msg *Message) (em *EmailMessage, err error) {
 		m.SetHeader("References", strings.Join(ref, " "))
 	}
 
-	m.SetHeader("X-Mailer", "Caliopen-"+b.Config.AppVersion)
-
 	//TODO: In-Reply-To header
 	m.SetHeader("Subject", msg.Subject)
 
 	// Handle if message is encrypted or not
 	features := *msg.Privacy_features
 	crypt_method, ok := features["message_encryption_method"]
-	log.Info("features ", msg.Privacy_features)
 	if !ok {
 		messages.SanitizeMessageBodies(msg)
 		if msg.Body_html != "" {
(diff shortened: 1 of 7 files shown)


src/backend/brokers/go.emails/email.go
```diff
@@ -52,6 +52,17 @@ func (b *EmailBroker) MarshalEmail(msg *Message) (em *EmailMessage, err error) {
 		},
 		Message: msg,
 	}
+	// sha256 internal message id to form external message id
```
sapiens-sapide: you could make use of `NewMessageId` func you created in the same pkg (and put this func within email.go instead of encrypt.go)

src/backend/brokers/go.emails/encrypt.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Create mail.Address list indexed by participant type
+func formatAddressList(msg *Message) map[string][]string {
+	// Create address headers
+	addr_fields := newAddressesFields()
+	for _, participant := range msg.Participants {
+		address := formatAddress(participant)
+		switch participant.Type {
+		case ParticipantFrom:
+			addr_fields["From"] = append(addr_fields["From"], address.String())
```
sapiens-sapide: addresses formatting is not consistent with MarshalEmail formatting rule : miss embedding participant.Label. 
instead, why not make use of `addr_fields` previously computed by MarshalEmail as it's a map[string]string ?

src/backend/brokers/go.emails/encrypt.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// References headers
+	mainHeader.Set("Message-ID", "<"+b.NewMessageId(em.Message.Message_id.Bytes())+">")
+
+	// Formatted multipart body
+	body, err := formatBody(msg, mainHeader)
+	if err != nil {
+		return err
+	}
+	em.Email.Raw = body
```
sapiens-sapide: this will be overwritten by MarshalEmail !

src/backend/brokers/go.emails/email.go
```diff
@@ -104,10 +119,9 @@ func (b *EmailBroker) MarshalEmail(msg *Message) (em *EmailMessage, err error) {
...
 	}
 
-	m.SetHeader("X-Mailer", "Caliopen-"+b.Config.AppVersion)
-
 	//TODO: In-Reply-To header
 	m.SetHeader("Subject", msg.Subject)
+
 	messages.SanitizeMessageBodies(msg)
 	if msg.Body_html != "" {
 		m.AddAlternative("text/html", msg.Body_html)
```
sapiens-sapide: do we still need to add alternative `text/plain` if body is encrypted ? May adding an `if is_encrypted` check

src/backend/brokers/go.emails/encrypt.go
```diff
@@ -0,0 +1,148 @@
...
+	log.Debug("Raw encrypted mail", em.Email.Raw)
+
+	json_rep, _ := EmailToJsonRep(em.Email.Raw.String())
+	em.Email_json = &json_rep
```
sapiens-sapide: this will be overwritten by MarshalEmail !

src/backend/configs/lmtp.yaml
```diff
@@ -59,7 +59,7 @@ LDAConfig:
   log_received_mails: true
 
   # outbound
-  out_topic: outboundSMTP                                # NATS topic to listen to
+  out_topic: outboundIMAP                                # NATS topic to listen to
```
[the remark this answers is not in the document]
gdchamal: Ok I understand better, APIv2 -> nats for imap worker that qualify the identity, if local forward to lmtp.

It's not optimal circuit, we will simplify it later. For the moment I revert this change and add a link from imap_worker to lmtpd service in docker compose stack

src/backend/brokers/go.emails/encrypt.go
gdchamal: nope I shunt the MarshalEmail process if it's encrypted to produce a full mail.

src/backend/brokers/go.emails/email.go
gdchamal: if message is encrypted we will not reach this part of code

src/backend/brokers/go.emails/encrypt.go
gdchamal: I use participant.Label when constructing the mail.Address structure, I think this formatting is more robust. And I can't use any structure of gomail.v2 package, so this is why I can't use same mechanism than for a non encrypted mail (use of mail.Message.FormatAddress)

src/backend/brokers/go.emails/email.go
gdchamal: done in #7525f865

src/backend/brokers/go.emails/encrypt.go
```diff
@@ -12,14 +12,33 @@ import (
...
+	"math/rand"
 	"net/mail"
 	"strings"
 	"time"
 )
 
-// TODO
+var boundaryChars = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789+:=?")
+
+func RandomString(n int) string {
```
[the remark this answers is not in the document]
gdchamal: I use this SO answer : https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go

I found it elegant with `rune` usage ;)

src/backend/brokers/go.emails/encrypt.go
```diff
@@ -0,0 +1,129 @@
...
+	part2Body := bytes.NewBuffer([]byte(msg.Body_plain))
+
+	entityVersion := &message.Entity{Header: part1, Body: part1Body}
+	entityBody := &message.Entity{Header: part2, Body: part2Body}
+
+	entities := make([]*message.Entity, 2)
+	entities[0] = entityVersion
+	entities[1] = entityBody
+
+	mp, err := message.NewMultipart(mainHeader, entities)
```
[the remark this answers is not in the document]
gdchamal: Nope references not accepted by message.NewMultipart method:

```
# github.com/CaliOpen/Caliopen/src/backend/brokers/go.emails
../brokers/go.emails/encrypt.go:125:46: cannot use &entities (type *[]*message.Entity) as type []*message.Entity in argument to message.NewMultipart
```

src/backend/brokers/go.emails/encrypt.go
```diff
@@ -0,0 +1,129 @@
...
+	part2.SetContentDescription("Caliopen PGP encrypted message")
+	part2.SetContentDisposition("inline", dispoParams)
+	part2Body := bytes.NewBuffer([]byte(msg.Body_plain))
+
+	entityVersion := &message.Entity{Header: part1, Body: part1Body}
+	entityBody := &message.Entity{Header: part2, Body: part2Body}
+
+	entities := make([]*message.Entity, 2)
+	entities[0] = entityVersion
+	entities[1] = entityBody
```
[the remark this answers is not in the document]
gdchamal: done in #fd4afde8

src/backend/brokers/go.emails/encrypt.go
sapiens-sapide: `addr_fields` is a map[string]string, not a gomail.v2 struct and it is already computed before the call to `MarshalEncryptedEmail`. You could call `MarshalEncryptedEmail` with one more params : 
`MarshalEncryptedEmail(msg *Message, em *EmailMessage, addr_fields map[string]string) error {` and add string from addr_fields map instead of `address.String()`

PR #1169, opened 2019-01-08
use RawButton to make sure the button's type is defined

iamdey:
+ fix click contact's title saves an empty contact


iamdey:
+1 if you have better name

PR #1174, opened 2019-01-10
[WIP] Refactoring/backend/naming and mq doc

sapiens-sapide:
This branch prepares workspace before refactoring identities workers (workers refactoring will be done in another branch).  
At this stage it adds documentation for our messaging system : look at [doc/specifications/messaging-system](https://github.com/CaliOpen/Caliopen/blob/refactoring/backend/naming-and-mq-doc/doc/specifications/messaging-system/index.md).  
In this document, you'll read a proposal for renaming workers related to remote identities management and protocols : 
- idpoller -> identities-worker
- caliopen_lmtpd -> smtp-bridge
- imapworker -> imap-bridge
- twitterworker -> twitter-bridge

I need your OK before continuing further more in renaming.  
In commit #3b1bc2f6 I already renamed `idpoller` to `identities-worker` to show you impact.


.drone.yml
```diff
@@ -345,12 +345,12 @@ pipeline:
...
 
-  build-idpoller-release:
+  build-identities-worker-release:
     group: release2
     image: plugins/docker
-    dockerfile: src/backend/Dockerfile.identity-poller
+    dockerfile: src/backend/Dockerfile.identities-worker
     context: /srv/caliopen/src/backend
-    repo: registry.caliopen.org/caliopen_identity_poller
+    repo: registry.caliopen.org/caliopen_identities-worker
```
PabloPie:
```suggestion
    repo: registry.caliopen.org/caliopen_identities_worker
```

Dockerfiles use dashes but Docker repositories use exclusively underscores.

devtools/kubernetes/deployments/identities-worker-deployment.yaml
```diff
@@ -25,5 +25,5 @@ spec:
         configMap:
           name: caliopen-config
           items:
-          - key: idpoller.yaml
-            path: idpoller.yaml
+          - key: identities-worker.yaml
+            path: identities-worker.yaml
```
PabloPie: Not very important, I think minikube support will be dropped...

gdchamal:
Why such huge renaming ? The poller poll and workers work as it's done now.

Except caliopen_lmtp that should be named eventually smtpworker I'm not fan of such renaming.

sapiens-sapide:
The poller doesn't 'poll', it's a daemon that continuously triggers jobs on its own and listens to nats orders (more like 'mq-worker').  
On the other side, protocols handlers are more like 'lmtpd' : they are waiting for jobs coming in, do the job (which is mainly IO with an external provider — thus the name 'bridge'), then return to 'idle' state.

gdchamal:
This renaming impact all the current deployment jobs and will make next deployments more difficult. And as we currently run on 2 distincts platforms and not only one (kubernetes cluster) that make such update painless, it's not the best moment for such change.

Can we wait ? It's not important for a running plaform from my point of view

sapiens-sapide:
OK, I remove the renaming because we're lacking time… But you know : there are two difficulties in computers : **naming** and **caching** ;-) 

I'll close this PR but keep the branch for the work done on messaging documentation.

PR #1178, opened 2019-01-12
Feature/backend/compute discussion hash

gdchamal:
This feature add a POST /participants/discussion API v1 route.

Given a list of participants (as found in a message entity) this API output such structure :

```
{"hash": "xxxxxxxxxxxxxxxxx", "discussion_id": "xxxx-xxxxx-xxxx-xxxx"}
```

Discussion_id can be null if no existing discussion with these participants is found


iamdey:
Not big deal, but looks like a filter of discussions with 1 or 0 results. 
The query string is a bit «trickier» than a POST body, some frameworks handle this out of the box:

`GET /api/discussions?participants[]=xxx&participants[]=yyy`

the best example is the `application/x-www-form-urlencoded`

gdchamal:
Merge as is, will see later if GET verb and querystring syntax is needed.

As now it's a specific route and not a filter on /discussions route

PR #1180, opened 2019-01-14
handle lost authentication, redirect signin

iamdey:
+ automatic signout if err 401 is caught
+ populate redirect after signout
+ fix request settings until the user is requiring an authenticated page


iamdey:
changelog

PR #1182, opened 2019-01-16
add progressbar & and fix download attachments

iamdey:
capture d ecran de 2019-01-16 17-07-01

### DIFF
--- a/src/frontend/web_application/src/scenes/Discussion/components/MessageAttachments/index.jsx
+++ b/src/frontend/web_application/src/scenes/Discussion/components/MessageAttachments/index.jsx
@@ -1,7 +1,10 @@
-import React, { Component } from 'react';
+import React, { Component, createRef } from 'react';
 import PropTypes from 'prop-types';
 import { withI18n } from '@lingui/react';
-import { Link, TextBlock, FileSize, Icon } from '../../../../components';
+import { CancelToken } from 'axios';
+import { Button, TextBlock, FileSize, Icon } from '../../../../components';
+import getClient from '../../../../services/api-client';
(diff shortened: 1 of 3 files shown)


src/frontend/web_application/src/scenes/Discussion/components/MessageAttachments/index.jsx
```diff
@@ -13,8 +16,104 @@ class MessageAttachments extends Component {
...
+
+  sendFileToBrowser = (index, { data, filename, mime }) => {
+    const blob = new Blob([data], { type: mime || 'application/octet-stream' });
+    const blobURL = window.URL.createObjectURL(blob);
+    const link = document.createElement('a');
+    link.style.display = 'none';
+    link.href = blobURL;
+    link.setAttribute('download', filename);
+    this.downloadRefs[index].ref.current.appendChild(link);
+    link.click();
```
MonsieurLanza: I'm clearly not a fan of this kind of hack… No other way to make DL happen ?

iamdey: I didn't find other way to send the file to the browser. cf. https://github.com/kennethjiang/js-file-download/blob/master/file-download.js

iamdey: … or may be https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download
I'll check that

iamdey: arf no, it's an API for WebExtensions

PR #1187, opened 2019-01-23
Fix/backend/remote identities workers

sapiens-sapide:


src/backend/protocols/go.imap/fetcher.go
```diff
@@ -57,28 +57,34 @@ func (f *Fetcher) SyncRemoteWithLocal(order IMAPorder) error {
...
 		}
 	}
 	// save syncing state in db to prevent concurrent sync
 	(*userIdentity).Infos["syncing"] = time.Now().Format(time.RFC3339)
-	f.Store.UpdateUserIdentity(userIdentity, map[string]interface{}{
+	err = f.Store.UpdateUserIdentity(userIdentity, map[string]interface{}{
 		"Infos": userIdentity.Infos,
 	})
+	if err != nil {
+		log.WithError(err).Infof("[SyncRemoteWithLocal] failed to update remote identity <%s> : <%s>", order.UserId, order.RemoteId)
```
gdchamal: an error is not an Info log level message

PR #1191, opened 2019-01-27
Feature/backend/configurable features

gdchamal:
Privacy features are typed values, they are stored in cassandra store and permit the marshall/unmarshall value correctly in both languages.


gdchamal:
Will be helpful for PI compute but also extensability

PR #1204, opened 2019-02-27
Feature/backend/participant lookup

gdchamal:
Add a participant_lookup store model to build a more robust discussion hash value.

A contact_id and an email address or other participant identifier, can be related to the a participant_id that will be use to compute discussion hash.


gdchamal:
[better notion of participant for lookup](https://trello.com/c/kvB6Dvd3/176-better-notion-of-participant-for-lookup)

src/backend/main/py.main/caliopen_main/message/parameters/message.py
```diff
@@ -57,15 +60,12 @@ class Options:
     @property
     def hash_participants(self):
         """Create an hash from participants addresses for global lookup."""
```
sapiens-sapide: hash from participant_id

src/backend/components/py.pi/caliopen_pi/qualifiers/base.py
```diff
@@ -63,11 +65,18 @@ def get_participant(self, message, participant):
...
+        p.participant_id = lookup.participant_id
         log.debug('Will lookup contact {} for user {}'.
                   format(participant.address, self.user.user_id))
         c = Contact.lookup(self.user, participant.address)
         if c:
             p.contact_ids = [c.contact_id]
+            ParticipantLookup.get_or_create(self.user,
+                                            c.contact_id,
+                                            'contact',
+                                            p.participant_id)
```
sapiens-sapide: algo is incomplete because it doesn't handle the case where a contact already exists for participant. In this case, algo must NOT create a new participant_id.

src/backend/defs/go-objects/participant.go
```diff
@@ -49,11 +56,41 @@ func (p *Participant) UnmarshalMap(input map[string]interface{}) error {
...
+	}
+	if identifier, ok := input["identifier"].(string); ok {
+		pl.Identifier = identifier
+	}
+	if type_, ok := input["type"].(string); ok {
+		pl.Type = type_
+	}
+	if part_id, ok := input["participant_id"].(gocql.UUID); ok {
+		pl.ParticipantId.UnmarshalBinary(part_id.Bytes())
+	}
```
sapiens-sapide: miss field DateInsert

src/backend/main/go.backends/store/cassandra/participant_lookup.go
```diff
@@ -0,0 +1,40 @@
...
+	"strings"
+)
+
+func (cb *CassandraBackend) RetrieveParticipantLookup(user_id, identifier, typ string) (lookup *ParticipantLookup, err error) {
+
+	// retrieve participant lookup
+	lookup = new(ParticipantLookup)
+	m := map[string]interface{}{}
+	clean_id := strings.ToLower(identifier)
+	q := cb.SessionQuery(`SELECT * FROM participant_lookup WHERE user_id = ? AND identifier = ? AND type = ?`, user_id, clean_id, typ)
```
sapiens-sapide: should we able to query on `user_id` and `identifier` only ? (`typ` optional)

src/backend/main/py.main/caliopen_main/message/core/participant_lookup.py
```diff
@@ -0,0 +1,35 @@
...
+    def create(cls, user, identifier, type):
+        """Create a new ``ParticipantLookup`` instance."""
+        date = datetime.utcnow()
+        return super(ParticipantLookup, cls).create(user=user,
+                                                    identifier=identifier,
+                                                    type=type,
+                                                    date_insert=date)
+
+    @classmethod
+    def get(cls, user, identifier, type):
```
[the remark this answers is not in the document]
gdchamal: Nope we expect a single result from a get, so all attributes composing the primary key must be given

src/backend/main/go.backends/store/cassandra/participant_lookup.go
gdchamal: We can query this way but it's a collection retrieved and must be a distinct function

PR #1208, opened 2019-02-27
Tests/backend/identities workers

sapiens-sapide:
Add a set of tests for imap worker, twitter worker and identities workers.


sapiens-sapide:
[Go tests for identities workers](https://trello.com/c/bBYlNsQq/183-go-tests-for-identities-workers)

PR #1211, opened 2019-03-01
Fix chronological order in discussion

MonsieurLanza:
Copy messages array to prevent accidental discussion reversal.


src/frontend/web_application/src/services/message/index.js
```diff
@@ -14,7 +14,8 @@ export const isParticipantUser = (participant, user) => {
   return (participant.contact_ids && participant.contact_ids.some(isUserContactId)) || false;
 };
 
-export const getLastMessage = messages => sortMessages(messages, true)[0];
+// Array.prototype.sort() is destructive, hence we need a copy of messages array.
```
iamdey: sortMessages should not mutate an argument. don't you think the fix must be in `sortMessages` so?

MonsieurLanza: You are right, let's do this.

PR #1213, opened 2019-03-05
PWA: install & cache bundle

iamdey:
+ fix bundle generation in assets/assets
+ fix i18n extract
+ add webpack plugins Offline & manifest


src/frontend/web_application/locale/fr/messages.json
```diff
@@ -575,11 +577,13 @@
...
   "take-a-tour.step.intro.content": "<0>Avec Caliopen, vous pouvez accéder à tous vos messages privés (Email, et plus à venir) via un seul login. </0><1>Maintenant, jetez un coup d'oeil à nos principales fonctionnalités, telles que la gestion unifiée des messages, la recherche intuitive et plus encore !</1><2>La première fois, le fil de discussion peut sembler vide, vous pouvez facilement <3>ajouter un compte externe</3> depuis le menu utilisateur</2>",
   "take-a-tour.step.intro.title": "Bienvenue!",
   "take-a-tour.step.search.content": "<0>Chaque recherche peut inclure des filtres. Toutes les données non chiffrées peuvent être recherchées.</0><1>Ici, vous pouvez rechercher tout ce qui se trouve dans vos messages et vos contacts.</1>",
   "take-a-tour.step.search.title": "Recherche intuitive",
-  "take-a-tour.step.user-menu.content": "<0>Mettez à jour les informations de votre compte.</1><2>Personnalisez votre application dans vos paramètres.</2>",
+  "take-a-tour.step.user-menu.content": "<0>Mettez à jour les informations de votre compte.</0><1>Personnalisez votre application dans vos paramètres.</1><2>Et connectez des fournisseurs tel que Gmail ou Twitter.</2>",
```
MonsieurLanza: Conflicts with https://github.com/CaliOpen/Caliopen/pull/1212

src/frontend/web_application/src/index.jsx
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { Router } from 'react-router-dom';
 import PiwikReactRouter from 'piwik-react-router';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
MonsieurLanza: Where is that dependency ? Local ?

iamdey: actually it's due to the not good enough project directory's structure:

This package is not required to run in production, because it is bundled for the client and the `index.jsx` file is only used by the client build.

It would be different if the client build has it's own `package.json`.

PR #1222, opened 2019-03-14
fix last message not visible

iamdey:
force fetch last messages

I forgot my own pattern: 
* `request<Resource>` force fetch to get up to date data and store it to redux (network first strategy)
* `get<Resource>` cache first strategy
* `fetch<Resource>`may be not related to redux (like password related requests)

Well, the best would be to write down this to a rfc.


MonsieurLanza:
> force fetch last messages
> 
> I forgot my own pattern:
> 
>     * `request<Resource>` force fetch to get up to date data and store it to redux (network first strategy)
> 
>     * `get<Resource>` cache first strategy
> 
>     * `fetch<Resource>`may be not related to redux (like password related requests)
> 
> 
> Well, the best would be to write down this to a rfc.

Indeed, so that others can know there is a pattern here. ^^

PR #1225, opened 2019-03-15
hide install button when user has authorized installation

iamdey:


src/frontend/web_application/src/layouts/Page/components/InstallButton/index.jsx
```diff
@@ -1,43 +1,63 @@
...
+    const { notifySuccess } = this.props;
+
+    defferedPrompt.prompt();
+    const { outcome } = await defferedPrompt.userChoice;
+
+    if (outcome === 'accepted') {
+      notifySuccess({
+        message: (<Trans id="pwa.feedback.install-success">Caliopen has been installed on your device.</Trans>),
+      });
+      this.setState({ isInstalled: true });
```
MonsieurLanza: I don't understand the logic here. What if the PWA is already installed ?

iamdey: if so, the defferedPrompt will be undefined because the event catched in InstallPromptConsumer will never be triggered.

agree that might not be crystal clear in some months

iamdey: I've added a comment

PR #1229, opened 2019-03-18
Clickable links in messages

MonsieurLanza:
This simple proposal uses ~https://github.com/alexcorvi/anchorme.js/tree/gh-pages/src~ linkifyjs to add clickable links to plain text messages. 

Could be used to make custom intern `mailto` links with some additions.

### DIFF
--- a/src/frontend/web_application/src/scenes/Discussion/components/MailMessage/index.jsx
+++ b/src/frontend/web_application/src/scenes/Discussion/components/MailMessage/index.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import Moment from 'react-moment';
 import { Trans, withI18n } from '@lingui/react';
 import classnames from 'classnames';
+import Linkify from 'linkifyjs/react';
 import { withScrollTarget } from '../../../../modules/scroll';
 import { withPush } from '../../../../modules/routing';
 import { ParticipantLabel } from '../../../../modules/message';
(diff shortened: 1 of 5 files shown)


src/frontend/web_application/src/scenes/Discussion/components/MailMessage/index.jsx
```diff
@@ -57,16 +58,11 @@ class MailMessage extends Component {
...
     if (!isLocked) {
-      if (!message.body_is_plain) {
-        return (
-          <TextBlock nowrap={false} className="s-mail-message__content" dangerouslySetInnerHTML={{ __html: message.body }} />
-        );
-      }
-
       return (
-        <TextBlock nowrap={false}><pre className="s-mail-message__content">{message.body}</pre></TextBlock>
+        <TextBlock nowrap={false} className="s-mail-message__content" dangerouslySetInnerHTML={{ __html: body }} />
```
iamdey: actually yes, I think it's too dangerous

iamdey: try to receive a plain text message w/ the following and see what happens:

```
<script>alert(1)</script>
```

MonsieurLanza: Yes I understand the problem.

MonsieurLanza:
Now uses https://soapbox.github.io/linkifyjs 

This library will escape HTML, except the links, it creates.

PR #1230, opened 2019-03-19
fix position of tags' dropdown

iamdey:
refactor a bit dropdown
+ eventual solution to calc position when no control ref
+ fix text align for next features buttons

### DIFF
--- a/src/frontend/web_application/src/modules/tags/components/TagsForm/presenter.jsx
+++ b/src/frontend/web_application/src/modules/tags/components/TagsForm/presenter.jsx
@@ -2,17 +2,13 @@ import React, { Component, createRef } from 'react';
 import PropTypes from 'prop-types';
 import { Trans } from '@lingui/react';
 import isequal from 'lodash.isequal';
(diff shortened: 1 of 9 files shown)


src/frontend/web_application/src/modules/tags/components/TagsForm/presenter.jsx
```diff
@@ -27,39 +23,53 @@ class TagsForm extends Component {
...
-    if (!isequal(nextProps.tags, this.props.tags)) {
-      this.setState(generateStateFromProps(nextProps));
+  componentDidUpdate(prevProps) {
+    const propNames = ['tags'];
+    const hasChanged = propNames.some(propName =>
+      !isequal(this.props[propName], prevProps[propName]));
+
+    if (hasChanged) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState(prevState => this.constructor.generateStateFromProps(this.props, prevState));
```
MonsieurLanza: :'(

PR #1234, opened 2019-03-21
Fix openpgp form

MonsieurLanza:
Accept only private key in user account.


iamdey:
the tests that are broken has been disabled in a PR don't know why it fails here

(may be rebase before merge)

PR #1236, opened 2019-03-21
[WIP] Some MIME multipart handling

MonsieurLanza:
Extracts `text/plain` parts of encrypted multipart MIME messages.

It handles quoted-printable encoding, and should handle base64 (yet not fully tested).

What concerns me is that it adds 3 dependencies : 
* https://github.com/ProtonMail/mimemessage.js
* https://github.com/mathiasbynens/utf8.js 
* https://github.com/mathiasbynens/quoted-printable

And we may need more in the future as we encounter more encodings.


iamdey:
two related bugs: 
* https://github.com/CaliOpen/Caliopen/issues/1237 the discussion_id is not saved
* https://github.com/CaliOpen/Caliopen/issues/1241 on draft init, the participants are not defined

PR #1243, opened 2019-03-26
refactor getIdentities and identitiesSelector in the module

iamdey:


iamdey:
iso functional (almost)

PR #1259, opened 2019-04-02
Refactor/frontend/update deps

iamdey:
update of webpack is coming but is a bit more complicated because of multiple bc break


iamdey:
another PR with more things updated https://github.com/CaliOpen/Caliopen/pull/1262

PR #1262, opened 2019-04-04
Refactor/frontend/update deps babel

iamdey:


iamdey:
patate PR, review by commit will be more efficient

PR #1271, opened 2019-04-18
Fix crash on contact creation

MonsieurLanza:
Not fixed yet. More exceptions are thrown.


iamdey:
the problem is not solved.
Does this fix enough or pragmatical ? I'm not sure.

for example if you reload the page directly on new contact, it will not fails.

I'm looking at the pbm too and it looks like the upgrade of redux has a BC Break on some API (`import { getFormValues } from 'redux-form';`)

PR #1280, opened 2019-04-25
Fix/backend/better errors on delivery

gdchamal:
Better error logging when deliverying a message.

Found logs on production that must be better processed


gdchamal:
[Bug delivery](https://trello.com/c/TTskwPa6/214-bug-delivery)

PR #1281, opened 2019-04-25
Fix/frontend/attachements visibility

iamdey:


MonsieurLanza:
Yes, but : encrypted message always have 2 empty attachments (encrypted.asc & "unknown"), due to the way the backend treats MIME envelope, which are not real attachments. So we should provide a way to not display those, or to fix backend.

iamdey:
I prefer the 2nd solution, fix the backend :D 
@sapiens-sapide

sapiens-sapide:
are these attachments « inline » ? If yes, is it the same issue that @iamdey encounter when testing inline images ?

MonsieurLanza:
encrypted.asc has `Content-Disposition: inline`. I am not sure about the other one.

PR #1298, opened 2019-05-03
Strip hardcoded values

MonsieurLanza:
Pass env vars to client, strip "alpha.caliopen.org" from client code. One ref remains in tests, but has no incidence on client.


src/frontend/web_application/server/ssr/ssr-middleware.js
```diff
@@ -46,6 +48,7 @@ export default (req, res) => {
       ...initialStateSettings,
       settings: getDefaultSettings(getUserLocales()[0]),
     },
+    instanceConfig: getConfig(),
```
iamdey: unused ?

MonsieurLanza: Unused. Removed.

Issue #1311, opened 2019-05-07
BSOD when replying to twitter message


MonsieurLanza:
(related to #1276, but should not crash)

When typing the very first char in response to a twitter message, and no identity is selected, I get a BSOD.

PR #1320, opened 2019-05-09
mistyped parentMessage property

### LINKED ISSUE #1276: auto selection of the identity for twitter
1: when receiving a twitter message, there are no identities selected by default

Capture d'écran de 2019-04-23 17-39-11

2: when sending a twitter message and there are no participants with an email, it should filter identities

Capture d'écran de 2019-04-23 17-47-17

**edit:** actually only emails are available and so the draft cannot be sent

iamdey:
when parentMessage is a Twitter DM, it select correctly the protocol

fix #1276


MonsieurLanza:
Does it fix #1311 too ? Should not happen, but…

iamdey:
it partially fix it

Issue #1312, opened 2019-05-07
Twitter : message duplication


MonsieurLanza:
Twitter messages are duplicated when sent from Caliopen, Both with different participants (see icon below).
My guess : message is created within Caliopen, then sent, then fetched from Twitter. 
![image](https://user-images.githubusercontent.com/4181623/57289841-73c58d80-70bc-11e9-94ef-8eed71fe567f.png)

PR #1325, opened 2019-05-09
Fix typo in contact's CleanupLookups

sapiens-sapide:
fix for issue #1312


sapiens-sapide:
it's not the right commit message ! /o\

Issue #1308, opened 2019-05-07
When creating a remote identity do not update contact if already present


gdchamal:
Each time we add the same remote identity, the contact is updated and the identity added even if it already exist.

https://mm.caliopen.org/api/v4/files/wqkzrtaftiy73m91ywoyne8dje/preview

gdchamal:
Do not update contact if address already exist

Issue #1295, opened 2019-05-03
error when adding a remote identity


iamdey:
* T1: I add my email address to my contact details
* T2: I add the remote identity corresponding to the previous address

it fails:

```
POST /api/v2/identities/remotes
{
  protocol: "email",
  status: "active",
  type: "remote",
  identifier: …,
}
---
{
	"errors": [{
		"code": 4,
		"message": "[CreateUserIdentity] failed to add identity \u003cdbfcbe33-993a-4087-acd4-4dec7e64862d\u003e to user \u003c0455b2b1-3da7-41e4-8c14-d066b0135e79\u003e's contact \u003c82c41474-1903-44cb-9183-0a29c97b6626\u003e",
		"name": ""
	}, {
		"code": 422,
		"message": "nil",
		"name": ""
	}]
}
```

but it does not fails completely, 

the remote is actually created and account fetched

PR #1329, opened 2019-05-10
Prevent to add remote identity to contact if exists

sapiens-sapide:
fix for issues #1295 and #1308


src/backend/main/go.main/facilities/REST/contacts.go
```diff
@@ -283,6 +290,13 @@ func addIdentityToContact(storeContact backends.ContactStorage, indexContact bac
 		}
 		updatedFields["Emails"] = newContact.Emails
 	case TwitterProtocol:
+		// prevent duplicate
+		for _, socialId := range contact.Identities {
+			if socialId.Type == TwitterProtocol && socialId.Name == identity.Identifier {
```
gdchamal: Can we be more generic and not only prevent duplicate for twitter ?

PR #1358, opened 2019-05-16
install cypress.io for e2e tests

iamdey:
+ add 2 tests
+ disable sauce-connect service
+ do not enable video & screenshot

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -80,6 +80,7 @@ pipeline:
   test-frontend:
     group: test
     image: cypress/base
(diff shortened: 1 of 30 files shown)


iamdey:
the results are very unstable for now, may be related the perf of the host.

iamdey:
It looks like https://github.com/smooth-code/jest-puppeteer could be a good alternative to cypress.

pro: made for chrome·ium, stability …
con: does not replace selenium and its support of multiple browsers

PR #1361, opened 2019-05-17
refactor default_locale detection

### LINKED ISSUE #1109: default locale is en_US and displayed French in settings
it happens when browser locale is en_US and because caliopen only knows en_EN, so the language selector is not good and displays French.

capture d ecran de 2018-11-23 11-41-33

To effectively change language to French:
1/ save to english or deutch
2/ save to French

iamdey:
+ fix bad locale saved when unknown
+ save default available lang when unsupported locale
+ change format of the locale from full normalized to simple laguage code (e.g. fr_BE -> fr)

fix #1109


iamdey:
@gdchamal the `default_locale` saved in settings will be a simple language code. 
Currently it is only by the i18n module of the frontend and support this format.

Do you see any other impact ?

iamdey:
as discussed I changed to save a full locale cf. https://github.com/CaliOpen/Caliopen/issues/1109#issuecomment-493481625

src/frontend/web_application/src/scenes/ApplicationSettings/components/InterfaceSettings/presenter.jsx
```diff
@@ -28,14 +27,14 @@ class InterfaceSettings extends PureComponent {
     const { i18n } = this.props;
     this.i18n = {
       fr_FR: i18n._('settings.interface.language.options.fr', null, { defaults: 'French' }),
```
MonsieurLanza: Shouldn't the language name be the only untranslated string ?

iamdey: :+1:  maybe something like «Français (French)»
I don't know if there is a RFC/bestpractice/lib… that can help

PR #1365, opened 2019-05-19
Feature/backend/import contact dedup

gdchamal:
We must deduplicate and filter contact informations when importing a vcard file.

Start building a logic around a dry run mode to report problems before they occurs.


gdchamal:
An apiv2 solution is preferred : https://github.com/CaliOpen/Caliopen/pull/1397

PR #1394, opened 2019-05-24
Contact consolidation

MonsieurLanza:
**NEEDS #1362**

While creating a new message, check all participants and add contact when it is missing, but available. Thus ensure encryption will not fail due to missing contact reference.

### DIFF
--- a/src/frontend/web_application/src/modules/draftMessage/actions/consolidateParticipants.js
+++ b/src/frontend/web_application/src/modules/draftMessage/actions/consolidateParticipants.js
@@ -1,0 +1,18 @@
+import { requestContactIdsForURI } from '../../../store/modules/contact';
+import { tryCatchAxiosAction } from '../../../services/api-client';
+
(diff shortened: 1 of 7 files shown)


src/frontend/web_application/src/modules/draftMessage/actions/consolidateParticipants.js
```diff
@@ -0,0 +1,18 @@
...
+const consolidateParticipant = participant => async dispatch => ({
+  ...participant,
+  contact_ids: [
+    ...(await tryCatchAxiosAction(() => dispatch(requestContactIdsForURI(
+      { protocol: participant.protocol, address: participant.address }
+    )))).contacts.map(contact => contact.contact_id),
+  ],
+});
+
+export const consolidateParticipants = ({ participants }) => async dispatch => Promise
```
iamdey: `async` not useful here

PR #1397, opened 2019-05-26
Feature/backend/import contact v2

gdchamal:
Move POST /imports route for vcard file processing to apiv2.

This way we benefit contact entry point deduplication. For the moment contacts are rejected when we try to create them during a vcard file import.

A better logic is needed (dry run, merge/forgot/etc strategy, ..)

Only vcard format 3.0 is parsed nowaday.

### DIFF
--- a/src/backend/interfaces/REST/go.server/operations/imports/import.go
+++ b/src/backend/interfaces/REST/go.server/operations/imports/import.go
@@ -1,0 +1,51 @@
+/*
+ * // Copyleft (ɔ) 2019 The Caliopen contributors.
+ * // Use of this source code is governed by a GNU AFFERO GENERAL PUBLIC
+ * // license (AGPL) that can be found in the LICENSE file.
+ */
+
+package imports
+
+import (
+	. "github.com/CaliOpen/Caliopen/src/backend/defs/go-objects"
+	"github.com/CaliOpen/Caliopen/src/backend/interfaces/REST/go.server/middlewares"
+	"github.com/CaliOpen/Caliopen/src/backend/interfaces/REST/go.server/operations"
+	"github.com/CaliOpen/Caliopen/src/backend/main/go.main"
+	"github.com/gin-gonic/gin"
+	swgErr "github.com/go-openapi/errors"
+	"net/http"
+)
+
+// ImportFile handles POST /imports and do logic depending on file mime type
+func ImportFile(ctx *gin.Context) {
+	userId, err := operations.NormalizeUUIDstring(ctx.MustGet("user_id").(string))
+
+	if err != nil {
+		e := swgErr.New(http.StatusUnprocessableEntity, err.Error())
+		http_middleware.ServeError(ctx.Writer, ctx.Request, e)
+		ctx.Abort()
+		return
+	}
+	shard_id := ctx.MustGet("shard_id").(string)
+	user_info := &UserInfo{User_id: userId, Shard_id: shard_id}
+
+	file, _, err := ctx.Request.FormFile("file")
+	if err != nil {
+		e := swgErr.New(http.StatusUnprocessableEntity, err.Error())
+		http_middleware.ServeError(ctx.Writer, ctx.Request, e)
+		ctx.Abort()
+		return
+	}
+
+	// TODO manage import switch file content-type / name
+	err = caliopen.Facilities.RESTfacility.ImportVcardFile(user_info, file)
+	if err != nil {
+		var e error
+		e = swgErr.New(http.StatusInternalServerError, err.Error())
+		http_middleware.ServeError(ctx.Writer, ctx.Request, e)
+		ctx.Abort()
+	} else {
+		ctx.Status(http.StatusOK)
+	}
+	return
+}
--- a/src/backend/interfaces/REST/go.server/operations/imports/import.go
+++ b/src/backend/interfaces/REST/go.server/operations/imports/import.go
@@ -41,7 +41,7 @@ func ImportFile(ctx *gin.Context) {
 	err = caliopen.Facilities.RESTfacility.ImportVcardFile(user_info, file)
 	if err != nil {
 		var e error
-		e = swgErr.New(http.StatusInternalServerError, err.Error())
+		e = swgErr.New(http.StatusUnprocessableEntity, err.Error())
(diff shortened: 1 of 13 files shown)


src/backend/main/go.main/contact/contact.go
```diff
@@ -0,0 +1,210 @@
...
+
+	//handle only first key found for now
+	if len(entitiesList) > 1 {
+		return nil, errors.New("More than one key found in payload")
+	}
+	return entitiesList[0], nil
+}
+
+// Parse KEY vcard field and transform to a PublicKey that belong to a contact
+func parseKey(field *vcard.Field, contact *objects.Contact) (*objects.PublicKey, error) {
```
sapiens-sapide: elsewhere in our stack, we use to return a nil pointer if func fails, instead of returning an empty or incomplete object. Here var `key` is return even if there were an error.

src/backend/main/go.main/contact/vcard.go
```diff
@@ -0,0 +1,49 @@
...
+		return card, nil
+	}
+}
+
+// Parse .vcf .vcard file, returting list of Card objects
+func ParseVcardFile(file io.Reader) ([]vcard.Card, error) {
+	cards := make([]vcard.Card, 0, 5)
+
+	scanner := bufio.NewScanner(file)
+	scanner.Split(bufio.ScanLines)
```
sapiens-sapide: instead of scanning each line until `END:VCARD` with `bufio.ScanLines` split func, then rebuild each line again,  you could have implemented your own split func that split on `VCARD` delimiters.

src/backend/main/go.main/facilities/REST/contacts.go
```diff
@@ -279,6 +281,41 @@ func (rest *RESTfacility) ContactExists(userID, contactID string) bool {
...
+	importErrors := make([]error, 0, len(vcards))
+	for _, card := range vcards {
+		contact, err := contact.FromVcard(info, card)
+		if err != nil {
+			log.Warn("[ImportVcardFile] Error during vcard transformation ", err)
+			importErrors = append(importErrors, err)
+		} else {
+			err = rest.CreateContact(info, contact)
+			if err != nil {
+				log.Warn("[ImportVcardFile] Create contact failed with error ", err)
```
sapiens-sapide: `rest.CreateContact` reports which uri is blocking within this contact in case of duplication with another contact. We could add a TODO here to handle this specific error and try to remove conflicting uri before calling again `rest.CreateContact`. It would improve the number of imported contacts.

src/backend/main/go.main/facilities/REST/contacts.go
```diff
@@ -279,6 +281,41 @@ func (rest *RESTfacility) ContactExists(userID, contactID string) bool {
 	return rest.store.ContactExists(userID, contactID)
 }
 
+// Process a vcard file and create related contacts
+func (rest *RESTfacility) ImportVcardFile(info *UserInfo, file io.Reader) error {
+	vcards, err := contact.ParseVcardFile(file)
```
sapiens-sapide: if `contact.ParseVcardFile` returns an empty array it's not reported as an error, it should ?

src/backend/interfaces/REST/go.server/operations/imports/import.go
```diff
@@ -0,0 +1,51 @@
...
+	file, _, err := ctx.Request.FormFile("file")
+	if err != nil {
+		e := swgErr.New(http.StatusUnprocessableEntity, err.Error())
+		http_middleware.ServeError(ctx.Writer, ctx.Request, e)
+		ctx.Abort()
+		return
+	}
+
+	// TODO manage import switch file content-type / name
+	err = caliopen.Facilities.RESTfacility.ImportVcardFile(user_info, file)
```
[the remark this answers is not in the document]
gdchamal: As discussed I'll raise an error if no vcard can be parsed in the file.

We do not have content-type information to ensure if it's a vcard, the filename is not a valid option

src/backend/main/go.main/contact/vcard.go
gdchamal: Yes you're right, my first experimentation around vcard decoding does not handle file with multiples entries. I did something not correctly. I'll simplify my implementation

src/backend/main/go.main/facilities/REST/contacts.go
gdchamal: done in #1a14ef7c3

src/backend/main/go.main/facilities/REST/contacts.go
gdchamal: These array of error will need further processing and first to be reported to user.

For the moment we reject all conflicting vcard

src/backend/main/go.main/contact/vcard.go
gdchamal: done in #984482dda

src/backend/main/go.main/contact/contact.go
gdchamal: done in #12bbebf81

src/backend/main/go.main/facilities/REST/contacts.go
```diff
@@ -279,6 +281,41 @@ func (rest *RESTfacility) ContactExists(userID, contactID string) bool {
...
+	if err != nil {
+		return err
+	}
+	log.Debug("[ImportVcardFile] Have parse ", len(vcards), " vcards")
+
+	importErrors := make([]error, 0, len(vcards))
+	for _, card := range vcards {
+		contact, err := contact.FromVcard(info, card)
```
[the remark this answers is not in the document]
gdchamal: done in #eeb8357e4

src/backend/main/go.main/contact/contact.go
```diff
@@ -202,9 +202,5 @@ func FromVcard(user *objects.UserInfo, card vcard.Card) (*objects.Contact, error
 		}
 	*/
 
-	// Compute a title if none found
```
sapiens-sapide: why do you remove `helpers.ComputeNewTitle(contact)` ? I was suggesting to just remove the `if` around it

gdchamal: This Title compute is done by CreateContact ;)

gdchamal:
I add email, phone or IM only if an identifier is not empty, see f04fd7445

PR #1411, opened 2019-05-29
Fix/backend/stop twitter account

sapiens-sapide:
Better handling of stopping twitter account handlers when they need to.

### DIFF
--- a/src/backend/protocols/go.twitter/worker.go
+++ b/src/backend/protocols/go.twitter/worker.go
@@ -182,7 +182,9 @@ func (worker *Worker) Start(throttling ...time.Duration) {
 		for msg := range worker.Desk {
 			switch msg.order {
 			case closeAccountOrder:
-				worker.RemoveAccountHandler(msg.account)
+				if msg.account != nil {
+					worker.RemoveAccountHandler(msg.account)
+				}
 			default:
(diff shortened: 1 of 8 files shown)


src/backend/protocols/go.twitter/worker.go
```diff
@@ -247,9 +269,11 @@ func (w *Worker) RegisterAccountHandler(accountHandler *AccountHandler) {
 }
 
 func (w *Worker) RemoveAccountHandler(accountHandler *AccountHandler) {
-	workerKey := accountHandler.userAccount.userID.String() + accountHandler.userAccount.remoteID.String()
-	w.WorkersGuard.Lock()
-	accountHandler.Stop(true)
-	delete(w.AccountHandlers, workerKey)
-	w.WorkersGuard.Unlock()
+	if accountHandler != nil {
```
gdchamal: We should not call this function with a nil accountHandler or it's not a normal situation if it can occurs and a warning log could be useful to see this situation

sapiens-sapide: Difficult to know when this could occur, but it occurred sometime during my test ! Indeed, there are many ways to stop accountHandler (closing chan, sending stop orders, suicide on errors, etc.)…

Anyway, I added a check before calling RemoveAccountHandler.

PR #1412, opened 2019-05-31
Clean user_name and recovery _email when deleting an user

gdchamal:
When an user delete its account, he should expect to be able to create a new account with same username and recovery email than previous.

Clean these related structures when deleting an user (set the user.date_deleted to not null).


src/backend/main/go.backends/store/cassandra/users.go
```diff
@@ -71,7 +71,23 @@ func (cb *CassandraBackend) UserByRecoveryEmail(email string) (user *User, err e
...
+	}
+	user, err := cb.RetrieveUser(user_id)
+	if err != nil {
+		return err
+	}
+	err = cb.SessionQuery(`DELETE from user_recovery_email WHERE recovery_email = ?`, user.RecoveryEmail).Exec()
+	if err != nil {
+		return err
+	}
+	err = cb.SessionQuery(`DELETE from user_name WHERE name = ?`, user.Name).Exec()
```
sapiens-sapide: I understand the point, but allowing to re-use a previous deleted username may allow a new user to receive messages previously addressed to the former username holder… I would prefer to add a `deleted` column to `user_name` table.  
It'd allow to automatically release previous username after time elapse, or to do it manually on user request if we agree.

gdchamal: I'll commit to remove deletion in user_name to get a retention period.

A script to delete correctly after a given period of time is soon ready (I needed it for data migration experiences). This script will clean user_name accordingly to the real delete schedule (a configurated value)

gdchamal: user_name no more deleted with ad9735f4e

PR #1418, opened 2019-06-04
Change send button icon

MonsieurLanza:
Change send button icon on advanced draft form for a more legible one (and consistent with quick draft) :
image

This is nothing, but it keeped bothering me for some time.

### DIFF
--- a/src/frontend/web_application/src/modules/draftMessage/components/DraftMessage/presenter.jsx
+++ b/src/frontend/web_application/src/modules/draftMessage/components/DraftMessage/presenter.jsx
@@ -647,11 +647,7 @@ class DraftMessage extends Component {
             disabled={!canSend}
           >
             {this.state.isSending && (<Spinner display="inline" theme="bright" />)}
-            {!this.state.isSending && (<Icon type="laptop" />)}
-            {' '}
-            ---
-            {' '}
-            <Icon type="user" />
+            {!this.state.isSending && (<Icon type="paper-plane" />)}
             {' '}
             <Trans id="draft-message.action.send">Send</Trans>
           </Button>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Move vcard file import route on apiv2 to use contact uniqueness and lookups principles
 - Re-enable import contacts via vcard file.
 - New messages notification is not displayed anymore, it is now automatically loaded.
+- Send button icon in advanced form.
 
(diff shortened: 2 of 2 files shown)


src/frontend/web_application/src/modules/draftMessage/components/DraftMessage/presenter.jsx
```diff
@@ -647,11 +647,8 @@ class DraftMessage extends Component {
             disabled={!canSend}
           >
             {this.state.isSending && (<Spinner display="inline" theme="bright" />)}
-            {!this.state.isSending && (<Icon type="laptop" />)}
-            {' '}
-            ---
-            {' '}
-            <Icon type="user" />
+            {this.state.isSending && ' '}
```
iamdey: `!this.state.isSending && …`

MonsieurLanza:
I asked once @judbd about it and says his first design was outdated, and we do not reflect the whole scheme of his pictos (which was pi based : context, transport, and contact based).

src/frontend/web_application/src/modules/draftMessage/components/DraftMessage/presenter.jsx
MonsieurLanza: This space is not meant as a placeholder for spinner but to actually add a space between spinner and icon.

Could not add it to the first condition, eslint complains.

judbd:
image

Best case scenario : the button reflects the state of the message and give contextual hint before sending if possible (non crypted chanel, suspicious address…)

If not possible: go for the paper plane.

Note: wtf is this beige? 😅

src/frontend/web_application/src/modules/draftMessage/components/DraftMessage/presenter.jsx
iamdey: oups I didn't read correctly

what about:
```
{this.state.isSending && (<Spinner display="inline" theme="bright" />)}
{!this.state.isSending && (<Icon type="paper-plane" />)}
```

The idea is to to not change the size of the button

MonsieurLanza:
> image
> 
> Best case scenario : the button reflects the state of the message and give contextual hint before sending if possible (non crypted chanel, suspicious address…)
> 
> If not possible: go for the paper plane.
> 
> Note: wtf is this beige? sweat_smile

The only information we are given is crypted / not crypted. No other hints, as PI is only calculated when message is effectively sent.

It is beige (well, yellow) as it is the color in caliopen security palette that reflect : this ain't safe. It can be greenish if message is encrypted. Mainly because difference between blue and greenish is not visible.

PR #1454, opened 2019-09-23
[WIP][frontend/encryption] Replaces `array.includes()` by custom func for compatibility.

MonsieurLanza:
May lead to unwanted participants key check failures on some older and/or Microsoft browsers, thus unencrypted mails.

Edit/WIP : found some more.


MonsieurLanza:
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/includes 
Should not be blocking… I close.

PR #1476, opened 2020-04-25
Frontend/upd deps

iamdey:


MonsieurLanza:
Can we upgrade node ? My latest attempts at it failed, but node 8 is getting old.

iamdey:
I did @MonsieurLanza in https://github.com/CaliOpen/Caliopen/pull/1470

PR #1501, opened 2021-02-12
Frontend/updates

iamdey:


src/frontend/web_application/src/modules/message/models/Message.ts
```diff
@@ -7,32 +7,53 @@ export class Message {
   constructor(props = {}) {
     Object.assign(this, props);
   }
+
```
MonsieurLanza, line 10: Why those space ? Feels like LF/CRLF conversion ?

MonsieurLanza (approved):

src/frontend/web_application/src/modules/message/models/Message.ts
iamdey: probably prettier

iamdey:
btw, this PR does not build due to ts errors

iamdey:
build fixed,

replace raw redux configure store by reduxjs/toolkit

src/frontend/web_application/src/modules/message/models/Message.ts
iamdey: I remember, it's autofix eslint

iamdey:
still something wrong at build, on live dev, the watcher restart in a loop, may be it's not blocking.

PR #1507, opened 2021-03-10
[frontend] fix import openpgp manager

iamdey:
refactor double declaration of `generateKey`


iamdey:
not fully verified yet

iamdey:
should be fine now

src/frontend/web_application/src/components/InputFileGroup/index.jsx
```diff
@@ -116,8 +117,8 @@ class InputFileGroup extends Component {
         {descr && <p>{descr}</p>}
 
         {this.state.files.length > 0 ? (
-          this.state.files.map((file) => (
-            <File file={file} onRemove={this.resetForm} />
+          this.state.files.map((file, i) => (
+            <File key={i} file={file} onRemove={this.resetForm} />
```
MonsieurLanza, line 121: No array index in key… Mais que fait `eslint` ? :p

src/frontend/web_application/src/scenes/UserSecurity/components/OpenPGPKey/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+                </Button>
+              )}
+            </div>
+          </>
+        )}
+      </div>
+
+      {keyDetails && (
+        <div className="m-openpgp-key__summary">
+          <span>{keyDetails?.userIds}</span>
```
MonsieurLanza, line 153: `userIds` are thrown as is, without separators, which is difficult to read. We may want to dive in a little bit to see how to display this in a better way.

src/frontend/web_application/src/components/InputFileGroup/index.jsx
iamdey: at least index is better than nothing :)

src/frontend/web_application/src/scenes/UserSecurity/components/OpenPGPKey/index.tsx
iamdey: fixed

PR #1515, opened 2021-04-15
Fork and patch go-mastodon library and small bugfix in mastodon worker

LaurentChemla:


LaurentChemla:
https://github.com/mattn/go-mastodon has been updated

PR #1522, opened 2021-05-08
[frontend] POC rtk-redux-query

iamdey:
- implement for tags
- manage progressive migration with legacy store and components

req #1521


iamdey:
**pro**

- looks nice and very well conceived 
- pretty stable
- abstract cache management and invalidation

**cons**

- still not sec / stable released
- low level actions not easy to manipulate and not well documented
- is it easy to use for newbies?
- a lot of abstraction


other thoughs?

iamdey:
release beta 3 is the last release before integrated to rtk and it is ok to start using it in prod: https://github.com/rtk-incubator/rtk-query/releases

PR #1534, opened 2021-07-27
[frontend] server: update iron, 

iamdey:
- install express types, 
- migrate some files to ts


iamdey:
still not fixed.

It's on the signup

iamdey:
at least the account is correctly created

iamdey:
That's better now

PR #1556, opened 2022-03-13
Create github-actions-demo.yml

iamdey:


gdchamal:
Ah OK. There was an outage with the whole platform down some weeks ago. May be the non working drone execution is related

PR #1590, opened 2023-05-03
[backend] fix empty discussion list

iamdey:
the sub aggregate term `discussion_id.keyword` might have been added by mistake

cf. 800388c47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Fixed
+
+- Empty discussion list.
+
 ## [0.26.0] 2023-04-04
 
 ### Changed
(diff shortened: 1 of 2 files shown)


gdchamal:
Hum I did an explicit commit to add this `.keyword` property on discussion_id for a performance issue. If it doesn't work I expect the ES schema to not be complete with correct definition of `discussion_id.keyword` to let ES index this property correctly.

iamdey:
Oh I supposed something like that. Well the problem is the es mapping was completely inconsistent and sometime old on some shards for an obscure reason and introduced err 500 on contact list. 
The shard migration command become unusable since the shard versioning system [has been changed](10948e1f98eb32be6af85b20d5e09ac3fff29c76). (So I disabled the command since it breaks the shards).
So I started to recreate all shards (manually dropped then recreated using `resync_shard` command) which does not use the mapping you expect. 
In the current state, I don't see where is this expected mapping:
- eventually create index: https://github.com/CaliOpen/Caliopen/blob/master/src/backend/main/py.main/caliopen_main/user/core/setups.py#L93-L96
- message mapping: https://github.com/CaliOpen/Caliopen/blob/master/src/backend/main/py.main/caliopen_main/message/store/message_index.py#L81

